PR #4, opened 2021-12-20
Added exponential timeout depending on the current round with a cap.

0xSasaPrsic:
Implementation of timeout function with a cap in order not to get stuck in RoundChangeState when the timeout is too large.

### DIFF
--- a/consensus.go
+++ b/consensus.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"math"
 	"os"
 	"reflect"
 	"time"
@@ -58,6 +57,7 @@ func WithTracer(t trace.Tracer) ConfigOption {
 
 const (
 	defaultTimeout = 2 * time.Second
+	maxTimeout     = 300 * time.Second
 )
 
 func DefaultConfig() *Config {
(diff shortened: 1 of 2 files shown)


consensus_test.go
```diff
@@ -465,8 +465,8 @@ func (m *mockPbft) emitMsg(msg *MessageReq) {
 
 func (m *mockPbft) addMessage(msg *MessageReq) {
 	// convert the address from the address pool
-	//from := m.pool.get(string(msg.From)).Address()
-	//msg.From = from
+	// from := m.pool.get(string(msg.From)).Address()
```
ferranbt, line 468: Why is this being commented?

consensus.go
```diff
@@ -683,16 +683,14 @@ func (p *Pbft) forceTimeout() {
...
-		timeout = inputTimeout[0]
-	}
-
+// exponentialTimeout calculates the timeout duration depending on the current round
+func (p *Pbft) exponentialTimeout() time.Duration {
+	timeout := defaultTimeout
 	round := p.state.view.Round
-	if round > 0 {
-		timeout += time.Duration(math.Pow(2, float64(round))) * time.Second
+	if round <= 8 {
```
ferranbt, line 690: Can you use `maxTimeout` here?

ferranbt (changes requested):

consensus_test.go
sprsic: This was already commented on in the main branch, and I did only formatting comments. Should I revert formatting?

consensus.go
0xSasaPrsic: Added a comment here about for <=8 as discussed.

PR #7, opened 2022-01-06
POS-98

0xSasaPrsic:
Implementation of tests for PBFT consensus.


consensus.go
```diff
@@ -600,25 +614,25 @@ func (p *Pbft) runRoundChangeState(ctx context.Context) {
...
 }
 
-func (p *Pbft) sendCommitMsg() {
-	p.gossip(MessageReq_Commit)
+func (p *Pbft) sendCommitMsg() error {
+	return p.gossip(MessageReq_Commit)
 }
 
-func (p *Pbft) gossip(typ MsgType) {
+func (p *Pbft) gossip(msgType MsgType) error {
```
ferranbt, line 633: So what should we do if there is a gossip error? What type of errors can we have in the gossip protocol? My understanding is that the Transport and Gossip function are async functions that should not fail to the user but handle any errors themselves.

Internally, a gossip transports enqueues all the possible messages from the Peer and does its best to broadcast them. I think we could even remove the error from the function altogether. Similar though applies to Sign. Thus, I think we can remove this errors from the gossip functions.

e2e/e2e_node_drop_test.go
```diff
@@ -10,8 +13,57 @@ func TestE2E_NodeDrop(t *testing.T) {
...
+func TestE2E_NetworkChurn(t *testing.T) {
+	rand.Seed(time.Now().Unix())
+	nodeCount := 20
+	maxFaulty := nodeCount/3 - 1
+	const prefix = "ptr_"
+	c := newPBFTCluster(t, "network_churn", "ptr", nodeCount)
+	c.Start()
+	runningNodeCount := nodeCount
+	// randomly stop nodes every 3 seconds
+	executeInTimerAndWait(3*time.Second, 30*time.Second, func(_ time.Duration) {
```
ferranbt, line 31: Do not make this an E2E test but a fuzz test. Create a different file called ```fuzz_network_churn_test.go``` and call the test ```TestFuzz_**``` instead.

e2e/framework.go
```diff
@@ -188,11 +188,25 @@ func (c *cluster) WaitForHeight(num uint64, timeout time.Duration, nodes ...[]st
 				return
 			}
 		case <-timer.C:
-			c.t.Fatal("timeout")
+			if throwPanic {
```
ferranbt, line 191: Why do you want to throwPanic here? What is the workflow of the tests that use this panic?

msg_queue_test.go
```diff
@@ -46,18 +44,41 @@ func TestMsgQueue_RoundChangeState(t *testing.T) {
...
+		m = newMsgQueue()
+		m.pushMessage(mockQueueMsg("A", MessageReq_RoundChange, ViewMsg(3, 1)))
+		assert.Nil(t, m.readMessage(RoundChangeState, ViewMsg(1, 1)))
+
+		m.pushMessage(mockQueueMsg("A", MessageReq_Commit, ViewMsg(3, 1)))
+		assert.Nil(t, m.readMessage(CommitState, ViewMsg(1, 1)))
+	}
+}
+
+func Test_msgToState(t *testing.T) {
```
ferranbt, line 65: I do not find it necessary to test that it panics.

e2e/e2e_unreliable_network_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"fmt"
+	"math/rand"
+	"strconv"
+	"testing"
+	"time"
+)
+
+func TestE2E_Unreliable_Network(t *testing.T) {
```
ferranbt, line 11: Is this a Fuzz or E2E test?

Makefile
```diff
@@ -3,6 +3,6 @@ test:
 	go test -v ./...
 
 e2e:
-	cd ./e2e && go test -v ./...
+	cd ./e2e && go test -timeout=50m -v ./...
```
ferranbt, line 6: If we only run the E2E tests in CI ```go test -v ./... -run TestE2e``` we should be able to remove this timeout. The Fuzz tests will not be tested on each PR but nightly most likely.

consensus.go
```diff
@@ -378,23 +394,22 @@ func (p *Pbft) runValidateState(ctx context.Context) { // start new round
...
+  
+	timeout := p.roundTimeout(p.state.view.Round)
 
-	timeout := p.exponentialTimeout()
 	for p.getState() == ValidateState {
 		_, span := p.tracer.Start(ctx, "ValidateState")
 
 		msg, ok := p.getNextMessage(span, timeout)
-		if !ok {
+		if !ok || msg == nil {
```
ferranbt, line 404: Please revert this change. I usually have a hard time myself with weird logic operands so I avoid them as much as possible.

consensus_test.go
```diff
@@ -11,287 +15,754 @@ import (
...
+type MockedTracer struct {
+	mock.Mock
+}
+
+func (m *MockedTracer) Start(ctx context.Context, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
+	args := m.Called(ctx, spanName, opts)
+	return args.Get(0).(context.Context), args.Get(1).(trace.Span)
+}
+
+func Test_InitializeConfigOptionsSet(t *testing.T) {
```
ferranbt, line 38: We do not need to do this test.

consensus_test.go
```diff
@@ -11,287 +15,754 @@ import (
...
+
+	assert.Equal(t, timeout, m.Pbft.config.Timeout)
+	assert.Equal(t, proposalTimeout, m.Pbft.config.ProposalTimeout)
+	assert.Equal(t, mockedTracer, m.Pbft.config.Tracer)
+
+	mockedTracer.AssertNotCalled(t, "Start")
+	mockedTracer.AssertExpectations(t)
+}
+
+func TestTransition_AcceptState(t *testing.T) {
```
ferranbt, line 73: Please revert this change. It is fine but when you are debugging or changing parts of the consensus and want to test specific functions, it is cumbersome to have to comment out this array instead of running ```go test -v ./... -run TestAcceptState_Something```.

consensus_test.go
```diff
@@ -11,287 +15,754 @@ import (
...
+		sc := sc
+		f := f
+		t.Run(sc, func(t *testing.T) {
+			t.Parallel()
+			f(t)
+		})
+	}
+}
+
+func TestTransition_RoundChangeState(t *testing.T) {
```
ferranbt, line 99: Same as in TestTransition_AcceptState.

consensus_test.go
```diff
@@ -11,287 +15,754 @@ import (
...
+		sc := sc
+		f := f
+		t.Run(sc, func(t *testing.T) {
+			t.Parallel()
+			f(t)
+		})
+	}
+}
+
+func TestTransition_ValidateState(t *testing.T) {
```
ferranbt, line 120: Same as in TestTransition_AcceptState.

ferranbt (changes requested):

ferranbt:
Please revert anything that is not test updates from consensus_test.go. There are so many changes in the way tests are run + variable changes that I cannot even figure out if there are any changes to the tests.

e2e/e2e_node_drop_test.go
0xSasaPrsic: :+1:

e2e/e2e_unreliable_network_test.go
0xSasaPrsic: It's fuzz test, will make adjustments as for previous fuzz test.

consensus.go
sprsic: sure, I will make adjustments.

e2e/framework.go
igorcrevar: This is added because we need somehow to detect timeout in tests. We found that it is easier to detect timeout in tests with panic/recover and we thought `WaitForHeight` is just a simple function inside e2e testing framework so introducing one more parameter is good enough solution. Otherwise the only other solution is to mock logs and catch t.Fatal("timeout")

consensus.go
0xSasaPrsic: sure, I will make adjustments.

Stefan-Ethernal: > So what should we do if there is a gossip error? 

We had this example on our mind when been thinking through it and thought maybe it is better to return an `error` object and caller can decide how to handle it. The question in the snippet below is: is it ok here to proceed with sending `prepare` messages if for some reason `sendPreprepareMsg` failed? If the answer is affirmative, then we should discuss and agree upon error handling strategy for each use case which can happen.
```
func (p *Pbft) runAcceptState(ctx context.Context) {
...
   // send the preprepare message
   p.sendPreprepareMsg()
   // send the prepare message since we are ready to move the state
   p.sendPrepareMsg()
...
}
```
On the other hand, we don't mind with only logging out errors, but thought that returning error object would be convenient.

> What type of errors can we have in the gossip protocol?

We don't know that yet, those are implementation details of validator seal signing (`Sign` function), gossiping on transport layer (`Gossip` function) etc. Namely each function which is being invoked within a 
`func (p *Pbft) gossip(msgType MsgType)` function.

0xSasaPrsic:
> Please revert anything that is not test updates from consensus_test.go. There are so many changes in the way tests are run + variable changes that I cannot even figure out if there are any changes to the tests.

Hey @ferranbt I made some changes, but there are some questions left in the comments, could you please take a look?

e2e/e2e_node_drop_test.go
ferranbt: Yes, everything that is not "deterministic" and has some random scenarios in the test itself should not be part of e2e_, all of those go to fuzz_xxx_test.go but keep on the same package.

e2e/e2e_unreliable_network_test.go
ferranbt: Move to fuzz_xxx_test.go and make it a fuzz test.

consensus.go
ferranbt: What are you going to do in the PBFT state machine if sending Preprepare fails?

Stefan-Ethernal: I was thinking maybe interrupting state machine loop and propagate an error back to the client immediately and it could decide about some retry logic, if it is important to send `prepare` message only after successful `pre-prepare` message sending. 
However, I got your point as well and seems this is not the case:
>  ...a gossip transports enqueues all the possible messages from the Peer and does its best to broadcast them.

which suggests that a node is doing its best to deliver all the messages to the peers and there is anyway `RoundChange` state when some essential condition is unsatisfied.

Just for the note, we have already reverted this change, just wanted to discuss around it a little bit more. :slightly_smiling_face:

e2e/e2e_unreliable_network_test.go
0xSasaPrsic: Fuzz tests moved to separate fail with appropriate test function name.

sprsic:
Closing this PR as it will be created in smaller chunks.

0xSasaPrsic:
Closing this PR as it is going to be created in several smaller chunks.

PR #11, opened 2022-01-17
Add e2e tests.

0xSasaPrsic:
Added more e2e tests.


e2e/e2e_issue_test.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (f *fsmInsertError) Insert(_ *pbft.SealedProposal) error {
+	return fmt.Errorf("insert error")
+}
+
+func createBackend(invalid bool) func(nn *node) pbft.Backend {
+	return func(nn *node) pbft.Backend {
+		if invalid {
+			switch rand.Intn(2) {
```
ferranbt, line 84: What is this? If this are not fuzz tests why do they have this random here?

e2e/e2e_issue_test.go
```diff
@@ -0,0 +1,94 @@
...
+	limit := int(math.Floor(nodesCnt * 2.0 / 3.0)) // + 1 removed because 2F+1 nodes is majority
+	for _, node := range c.Nodes() {
+		node.StartWithBackendFactory(createBackend(node.name < "prt_"+strconv.Itoa(limit)))
+	}
+	defer c.Stop()
+	names := generateNodeNames(limit, nodesCnt, "prt_")
+	err := c.WaitForHeight(3, 1*time.Minute, names)
+	if err == nil {
+		t.Fatal("Height reached for minority of nodes")
+	}
```
ferranbt, line 45: Restart the partition and wait for them to sync up again.

e2e/e2e_issue_test.go
```diff
@@ -0,0 +1,94 @@
...
+	hook := newPartitionTransport(300 * time.Millisecond)
+
+	c := newPBFTCluster(t, "majority_partition", "prt", nodesCnt, hook)
+	limit := int(math.Floor(nodesCnt*2.0/3.0)) + 1 // 2F+1 nodes can Validate
+	for _, node := range c.Nodes() {
+		node.StartWithBackendFactory(createBackend(node.name >= "prt_"+strconv.Itoa(limit)))
+	}
+	defer c.Stop()
+	names := generateNodeNames(0, limit, "prt_")
+	err := c.WaitForHeight(4, 1*time.Minute, names)
```
ferranbt, line 25: Use the assert library in the package. Import it if need it.

ferranbt (changes requested):

e2e/e2e_issue_test.go
0xSasaPrsic: It just creates a backend with "bad" behaviour, node that always error on validation or node that always return error on insert. It just picks randomly one of these 2 implementations. We can use only one implementation and remove random if needed.

e2e/e2e_issue_test.go
0xSasaPrsic: ok, bad nodes will be restarted with correct backend and wait for them to sync with the network.

e2e/e2e_node_drop_test.go
```diff
@@ -3,15 +3,18 @@ package e2e
...
 )
 
 func TestE2E_NodeDrop(t *testing.T) {
 	c := newPBFTCluster(t, "node_drop", "ptr", 5)
 	c.Start()
-
 	// wait for two heights and stop node 1
-	c.WaitForHeight(2, 1*time.Minute)
+	err := c.WaitForHeight(2, 1*time.Minute)
+	assert.NoError(t, err, "Error should not be returned.")
```
ferranbt, line 15: This does not need to be that verbose every time. Just doing
```
err := ...
assert.NoError(t, err)
```
should work.

e2e/e2e_partition_test.go
```diff
@@ -1,27 +1,132 @@
...
+}
+
+func TestE2E_Partition_MajorityCantValidate(t *testing.T) {
+	const nodesCnt = 7 // N=3F + 1, F = 2
+	hook := newPartitionTransport(300 * time.Millisecond)
+
+	c := newPBFTCluster(t, "majority_partition", "prt", nodesCnt, hook)
+	limit := int(math.Floor(nodesCnt * 2.0 / 3.0)) // + 1 removed because 2F+1 nodes is majority
+	for _, node := range c.Nodes() {
+		node.StartWithBackendFactory(createBackend(node.name < "prt_"+strconv.Itoa(limit)))
```
ferranbt, line 80: Ok, so you are using a `faulty` backend with `createBackend` to simulate that only a majority of nodes are available? If the backend is faulty, no blocks are validated. I like that we support this but the API to introduce this error is different from the partition one.

For partitions we do: 
```
hook.Partition([]string{}...)
```
And to drop a node we do:
```
c.StopNode("ptr_0")
```
The errors are introduced with some high level API either from the hooks (communication transport) or the cluster object (node errors).

The standard way to do this would be something like this:
```
func (c *cluster) FailNode(name string) {
  c.nodes[name].setValidationFails(true)
}

type fsm struct {
  validationFails bool
}

func (f *fsm) setValidationFails(v bool) {
  f.validationFails = v
}

func (f *fsm) Validate(proposal []byte) error {
  if f.validationFails {
    return fmt.Errorf("error")
  }
  return nil
}
```

ferranbt (changes requested):

e2e/e2e_partition_test.go
0xSasaPrsic: Changes made accordingly.

ferranbt (approved):

PR #12, opened 2022-01-17
POS-131: Pbft consensus unit tests

Stefan-Ethernal:
Implemented remaining unit tests for:
`state.go`,
`msg_queue.go`,
`consensus.go`.

Other than this, `consensus.go` has undergone several small changes:
- Implemented timeout mocking mechanism.
- Removed `setSequence` invocation on `runCommitState`, since sequence is set to block height from the outside.

### DIFF
--- a/consensus_test.go
+++ b/consensus_test.go
@@ -15,25 +15,13 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
-	"go.opentelemetry.io/otel/trace"
 )
 
 var (
 	mockProposal  = []byte{0x1, 0x2, 0x3}
 	mockProposal1 = []byte{0x1, 0x2, 0x3, 0x4}
 )
-
-// MockedTracer is a mocked object that implements Tracer interface
-type MockedTracer struct {
-	mock.Mock
-}
-
-func (m *MockedTracer) Start(ctx context.Context, spanName string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
-	args := m.Called(ctx, spanName, opts)
-	return args.Get(0).(context.Context), args.Get(1).(trace.Span)
-}
 
 func TestTransition_AcceptState_ToSyncState(t *testing.T) {
 	// we are in AcceptState and we are not in the validators list
--- a/consensus_test.go
+++ b/consensus_test.go
@@ -35,15 +35,15 @@ func (m *MockedTracer) Start(ctx context.Context, spanName string, opts ...trace
 	return args.Get(0).(context.Context), args.Get(1).(trace.Span)
 }
 
-func TestTransition_AcceptState_SyncStateTransition(t *testing.T) {
-	// we are in AcceptState, and we are not in the validators list
+func TestTransition_AcceptState_ToSyncState(t *testing.T) {
+	// we are in AcceptState and we are not in the validators list
 	// means that we have been removed as validator, move to sync state
-	m := newMockPbft(t, []string{"A", "B", "C", "D"}, "")
-	m.setState(AcceptState)
+	i := newMockPbft(t, []string{"A", "B", "C", "D"}, "")
+	i.setState(AcceptState)
 
-	m.runCycle(m.ctx)
+	i.runCycle(context.Background())
 
-	m.expect(expectResult{
+	i.expect(expectResult{
 		sequence: 1,
 		state:    SyncState,
 	})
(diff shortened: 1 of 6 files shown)


state_test.go
```diff
@@ -24,68 +71,307 @@ func TestState_FaultyNodes(t *testing.T) {
...
+		{5, 2},
+		{6, 2},
+		{7, 4},
+		{8, 4},
+		{9, 4},
+		{10, 6},
+		{100, 66},
+	}
+	for _, c := range cases {
+		validatorNodeIds := generateValidatorNodes(c.TotalNodesCount, "validator")
```
ferranbt, line 104: Do you ever need the result of this function `generateValidatorNodes`? It seems just an extra useless step.

state_test.go
```diff
@@ -3,17 +3,64 @@ package pbft
...
+}
+
+// Helper function which enables creation of MessageReq.
+func createMessage(sender string, messageType MsgType, round ...uint64) *MessageReq {
+	return createMessagePoolSender(sender, messageType, nil, round...)
+}
+
+// Helper function which enables creation of MessageReq.
+// Note: sender needs to be seeded to the pool before invoking, otherwise senderId will be set to provided sender parameter.
+func createMessagePoolSender(sender string, messageType MsgType, pool *testerAccountPool, round ...uint64) *MessageReq {
```
ferranbt, line 38: Add this as part of `testerAccountPool`?
```
func (t *testerAccountPool) createMessagePoolSender(sender string, typ MsgType, round ...uint64) ...
```

state_test.go
```diff
@@ -24,68 +71,307 @@ func TestState_FaultyNodes(t *testing.T) {
...
+	assert.Empty(t, s.prepared)
+	assert.Empty(t, s.committed)
+}
+
+func TestState_addPrepared(t *testing.T) {
+	s := newState()
+	validatorIds := []string{"validator_1", "validator_2"}
+	s.validators = newMockValidatorSet(validatorIds)
+
+	s.addPrepared(createMessage(validatorIds[0], MessageReq_Commit))
```
ferranbt, line 244: In most of the tests you do use ``validatorIds[0]`` to refer to one of the validators. Personally, this is a bit confusing for me because I have to double check the id of the validators all the time.

I would prefer to use simple names (A, B, C...) and then writing down the code directly.

state_test.go
```diff
@@ -24,68 +71,307 @@ func TestState_FaultyNodes(t *testing.T) {
...
+func TestState_GetSequence(t *testing.T) {
+	s := newState()
+	s.view = &View{Sequence: 3, Round: 0}
+	assert.True(t, s.GetSequence() == 3)
+}
+
+func TestState_getCommittedSeals(t *testing.T) {
+	s := newState()
+	validatorIds := []string{"A", "B"}
+	s.validators = newMockValidatorSet(validatorIds)
```
ferranbt, line 304: Do not be afraid to create a pool object here only to get the validator set. It is convenient.

ferranbt (changes requested):
`consensus_test.go` has too many changes to review. Could you revert:
- Context in runCycle.
- Rename of test files.
- `newMockPbft` variable rewrite.

state_test.go
Stefan-Ethernal: Result of function is added just because of readability.

Do you have this change in mind
convertToMockValidatorSet(generateValidatorNodes(c.TotalNodesCount, "validator"))
or there is more elegant way to rewrite that part? TBH, not sure whether we can avoid having generateValidatorNodes function, although I agree that it is pretty dummy.

Stefan-Ethernal: Second approach would be to instantiate `testerAccountPool` and to delete `generateValidatorNodes`. Although it could be overkill, could be more elegant way in terms of code brevity. :slightly_smiling_face:

state_test.go
Stefan-Ethernal: Have changed it, although it seems like it is slower than creating mock validator set, as I have pointed out in the https://github.com/0xPolygon/pbft-consensus/pull/12#discussion_r786076784.

consensus_test.go
```diff
@@ -11,89 +15,24 @@ import (
...
-	i.expect(expectResult{
-		sequence:    1,
-		state:       ValidateState,
-		prepareMsgs: 3,
-		commitMsgs:  1, // A commit message
-		locked:      true,
-		outgoing:    1, // A commit message
-	})
+// MockedTracer is a mocked object that implements Tracer interface
+type MockedTracer struct {
```
ferranbt, line 29: I think we do not need this right? It does not panic AFAIK when it is not set.

ferranbt (changes requested):
Changes look good but please split this PR in:
- state tests.
- consensus tests.
- random timeout.

consensus_test.go
Stefan-Ethernal: Indeed, will remove it.

Stefan-Ethernal:
Closing this PR, as changes in it will be subjects of some smaller PRs.

PR #13, opened 2022-01-18
POS-141: Minor code cleanup

Stefan-Ethernal:
Miscellaneous code cleanup and minor tweaks such as:

tidy go mod dependencies, 

variables renaming, 

log messages format fix, 

copying seal field in state.go.

### DIFF
--- a/state.go
+++ b/state.go
@@ -50,8 +50,8 @@ type MessageReq struct {
 	Proposal []byte
 }
 
-func (m *MessageReq) SetProposal(b []byte) {
-	m.Proposal = append([]byte{}, b...)
+func (m *MessageReq) SetProposal(proposal []byte) {
+	m.Proposal = append([]byte{}, proposal...)
 }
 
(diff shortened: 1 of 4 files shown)


ferranbt (dismissed):

state.go
```diff
@@ -61,7 +61,11 @@ func (m *MessageReq) Copy() *MessageReq {
 		mm.View = m.View.Copy()
 	}
 	if m.Proposal != nil {
-		mm.Proposal = append([]byte{}, m.Proposal...)
+		mm.SetProposal(m.Proposal)
+	}
+	// FIXME: Check if Seal should be copied as well or instantiated an empty slice instead?
```
Stefan-Ethernal, line 66: @ferranbt Before merging this commit onto main, do you agree with copying `Seal` as well, since I wasn't sure if it should be empty slice or nil maybe.

ferranbt: Yes, Seal is also copied.

ferranbt (approved):

PR #15, opened 2022-01-18
POS-147: Consensus unit tests

Stefan-Ethernal:
Implemented unit tests for PolyBFT consensus algorithm.
Removed incrementing sequence number during `CommitState` -> `DoneState` transition.

Note that some of unit tests are skipped until POS-150 gets merged to the main branch.

### DIFF
--- a/consensus_test.go
+++ b/consensus_test.go
@@ -1,7 +1,6 @@
 package pbft
 
 import (
-	"bytes"
 	"container/heap"
 	"context"
 	"crypto/sha1"
@@ -130,23 +129,18 @@ func TestTransition_AcceptState_Validator_VerifyFails(t *testing.T) {
 	})
 }
 
+// Test that if build proposal fails, state machine will change state from AcceptState to RoundChangeState.
 func TestTransition_AcceptState_Proposer_FailedBuildProposal(t *testing.T) {
 	buildProposalFunc := func() (*Proposal, error) {
 		return nil, errors.New("failed to build a proposal")
 	}
 
 	validatorIds := []string{"A", "B", "C"}
-	backend := newMockBackend(validatorIds, nil, buildProposalFunc, nil, nil)
+	backend := newMockBackend(validatorIds, nil).HookBuildProposalHandler(buildProposalFunc)
 
 	m := newMockPbft(t, validatorIds, "A", backend)
 	m.state.view = ViewMsg(1, 0)
 	m.setState(AcceptState)
-
-	var buf bytes.Buffer
-	m.logger.SetOutput(&buf)
-	defer func() {
-		m.logger.SetOutput(getDefaultLoggerOutput())
-	}()
 
 	// Prepare messages
 	m.emitMsg(&MessageReq{
@@ -174,37 +168,27 @@ func TestTransition_AcceptState_Proposer_FailedBuildProposal(t *testing.T) {
 		}
 	}()
 
-	m.Run(m.ctx)
-	m.logger.Println(buf.String())
-
-	assert.Contains(t, buf.String(), "[ERROR] failed to build proposal", "build proposal did not fail")
+	m.runCycle(m.ctx)
 	assert.True(t, m.IsState(RoundChangeState))
 }
 
+// Run state machine from AcceptState, proposer node.
+// Artificially induce state machine cancellation and check whether state machine is still in AcceptState.
 func TestTransition_AcceptState_Proposer_Cancellation(t *testing.T) {
-	// We are in the AcceptState, we are the proposer.
-	// Cancellation occurs, so we are checking whether we are still in AcceptState.
-	m := newMockPbft(t, []string{"A", "B", "C", "D"}, "A")
-	m.setState(AcceptState)
-	m.setProposal(&Proposal{
-		Data: mockProposal,
-		Time: time.Now().Add(time.Second),
-	})
-
-	go func() {
-		m.cancelFn()
-	}()
-
-	m.runCycle(m.ctx)
-	assert.True(t, m.IsState(AcceptState))
+	testAcceptState_Cancellation(t, true)
 }
 
+// Run state machine from AcceptState, non-proposer node.
+// Artificially induce state machine cancellation and check whether state machine is still in the AcceptState.
 func TestTransition_AcceptState_NonProposer_Cancellation(t *testing.T) {
-	t.Skip("Test lasts too long, because it waits for timeout to elapse since message queues are empty. Mocking timeout calculation need to be established.")
-	// We are in the AcceptState, we are non proposer node.
-	// Cancellation occurs, so we are checking whether we are still in AcceptState.
+	testAcceptState_Cancellation(t, false)
+}
+
+func testAcceptState_Cancellation(t *testing.T, isProposerNode bool) {
 	m := newMockPbft(t, []string{"A", "B", "C", "D"}, "D")
-	m.Pbft.state.proposer = "A"
+	if !isProposerNode {
+		m.Pbft.state.proposer = "A"
+	}
 
 	m.setState(AcceptState)
 	m.setProposal(&Proposal{
@@ -217,7 +201,7 @@ func TestTransition_AcceptState_NonProposer_Cancellation(t *testing.T) {
 	}()
 
 	m.runCycle(m.ctx)
-	assert.True(t, m.IsState(AcceptState))
+	assert.True(t, m.IsState(AcceptState), "Expected to be in %s, but was in %s", AcceptState, m.getState())
 }
 
 func TestTransition_AcceptState_Validator_ProposerInvalid(t *testing.T) {
@@ -244,7 +228,7 @@ func TestTransition_AcceptState_Validator_ProposerInvalid(t *testing.T) {
 }
 
 func TestTransition_AcceptState_Validator_LockWrong(t *testing.T) {
-	// we are a validator and have a locked state in 'proposal1'.
+	// We are a validator and have a locked state in 'proposal1'.
 	// We receive an invalid proposal 'proposal2' with different data.
 
 	i := newMockPbft(t, []string{"A", "B", "C"}, "B")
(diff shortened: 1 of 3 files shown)


consensus_test.go
```diff
@@ -11,90 +15,13 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
-func TestTransition_ValidateState_Prepare(t *testing.T) {
```
ferranbt, line 16: Have you removed this tests?

consensus_test.go
```diff
@@ -208,6 +130,96 @@ func TestTransition_AcceptState_Validator_VerifyFails(t *testing.T) {
...
+	}
+
+	validatorIds := []string{"A", "B", "C"}
+	backend := newMockBackend(validatorIds, nil, buildProposalFunc, nil, nil)
+
+	m := newMockPbft(t, validatorIds, "A", backend)
+	m.state.view = ViewMsg(1, 0)
+	m.setState(AcceptState)
+
+	var buf bytes.Buffer
```
ferranbt, line 145: Remove this, this logger should be abstracted away in the mock pbft. I do not want to put 5 lines on each test.

consensus_test.go
```diff
@@ -208,6 +130,96 @@ func TestTransition_AcceptState_Validator_VerifyFails(t *testing.T) {
...
+	go func() {
+		for {
+			if m.GetState() == RoundChangeState {
+				m.cancelFn()
+				return
+			}
+		}
+	}()
+
+	m.Run(m.ctx)
```
ferranbt, line 177: Do not use `Run` for this transition tests since it is a long live running function, use `runCycle` instead. Most likely you can remove the `go func()` as well.

consensus_test.go
```diff
@@ -208,6 +130,96 @@ func TestTransition_AcceptState_Validator_VerifyFails(t *testing.T) {
...
+	m.logger.Println(buf.String())
+
+	assert.Contains(t, buf.String(), "[ERROR] failed to build proposal", "build proposal did not fail")
+	assert.True(t, m.IsState(RoundChangeState))
+}
+
+func TestTransition_AcceptState_Proposer_Cancellation(t *testing.T) {
+	// We are in the AcceptState, we are the proposer.
+	// Cancellation occurs, so we are checking whether we are still in AcceptState.
```
ferranbt, line 186: What does cancellation occurs means?

consensus_test.go
```diff
@@ -291,6 +303,75 @@ func TestTransition_AcceptState_Validator_LockCorrect(t *testing.T) {
...
+	go func() {
+		for {
+			if m.GetState() == RoundChangeState {
+				m.cancelFn()
+				return
+			}
+		}
+	}()
+
+	m.Run(m.ctx)
```
ferranbt, line 355: Same comment as before for the `Run` function and the logger.

consensus_test.go
```diff
@@ -291,6 +303,75 @@ func TestTransition_AcceptState_Validator_LockCorrect(t *testing.T) {
 	})
 }
 
+func TestTransition_AcceptState_Validate_ProposalFail(t *testing.T) {
+	validateProposalFunc := func(proposal []byte) error {
```
ferranbt, line 307: Please add a description for the test.

consensus_test.go
```diff
@@ -291,6 +303,75 @@ func TestTransition_AcceptState_Validator_LockCorrect(t *testing.T) {
...
+
+	m.Run(m.ctx)
+	t.Log(buf.String())
+
+	assert.Contains(t, buf.String(), "[ERROR] failed to validate proposal", "validate proposal did not fail")
+	assert.True(t, m.IsState(RoundChangeState))
+}
+
+func TestTransition_AcceptState_NonValidatorNode(t *testing.T) {
+	// Node sending a messages isn't among validator set, so state machine should set state to SyncState
```
ferranbt, line 363: By "Node sending a messages" I think you mean the local node running pbft.

consensus_test.go
```diff
@@ -444,6 +525,294 @@ func TestTransition_RoundChangeState_MaxRound(t *testing.T) {
...
+		state:       CommitState,
+		prepareMsgs: 3,
+		commitMsgs:  3, // Commit messages (A proposer sent commit via state machine loop, C and D sent commit via emit message)
+		locked:      true,
+		outgoing:    1, // A commit message
+	})
+}
+
+func TestTransition_ValidateState_MoveToRoundChangeState(t *testing.T) {
+	t.Skip("Test lasts too long, because it waits for timeout to elapse since message queues are empty. Mocking timeout calculation need to be established.")
```
ferranbt, line 600: If it works, please remove the Skip and let CI run longer for a couple of PRs.

consensus_test.go
```diff
@@ -444,6 +525,294 @@ func TestTransition_RoundChangeState_MaxRound(t *testing.T) {
...
+	go func() {
+		close(ch)
+		for {
+			if m.getState() == AcceptState {
+				m.cancelFn()
+				return
+			}
+		}
+	}()
+	<-ch
```
ferranbt, line 758: What is this channel doing here?

consensus_test.go
```diff
@@ -444,6 +525,294 @@ func TestTransition_RoundChangeState_MaxRound(t *testing.T) {
...
+}
+
+func TestGossip_Failed(t *testing.T) {
+	m := newMockPbft(t, []string{"A", "B"}, "A")
+	m.gossipFn = func(msg *MessageReq) error {
+		return errors.New("deliberate failure")
+	}
+
+	var buf bytes.Buffer
+	m.logger.SetOutput(&buf)
```
ferranbt, line 787: Logger.

consensus_test.go
```diff
@@ -522,12 +902,34 @@ func newMockPbft(t *testing.T, accounts []string, account string) *mockPbft {
...
+	if os.Getenv("SILENT") == "true" {
+		return ioutil.Discard
+	}
+	return os.Stdout
+}
+
+func newMockBackend(
+	validatorIds []string,
+	mockPbft *mockPbft,
+	buildProposal buildProposalDelegate,
```
ferranbt, line 915: These callbacks are cumbersome to read and you have to use `nil` in much of the definitions of `newMockBackend`. We might use this pattern instead:

```
func (m *mockBackend) WithBuildProposal(buildProposal buildProposalDelegate) *mockBackend {
    m.buildProposal  = buildProposal
}

func TestXX(t *testing.T) {
   m := newmockBackend([]string{}, ..).WithBuildProposal(func...)
}
```

ferranbt (changes requested):

consensus_test.go
Stefan-Ethernal: `TestTransition_ValidateState_Prepare` is renamed to `TestTransition_ValidateState_CommitState` and slightly changed in order to satisfy conditions to move from Validate to Commit state.

On the other hand, `TestTransition_ValidateState_CommitFastTrack` is removed completely, since `TestTransition_ValidateState_CommitState` already tests `ValidateState` -> `CommitState` state transition.

consensus_test.go
Stefan-Ethernal: Ok, agreed. We just wanted to enforce assertions in order to be sure that expected part of code executed.

consensus_test.go
Stefan-Ethernal: Generally speaking, intention was to cover all the branches in the code base with unit tests, although this maybe isn't real use case.
This test specifically covers `p.ctx.Done()` channel from the snippet bellow:

```go
func (p *Pbft) runAcceptState(ctx context.Context) {
...
if isProposer {
...
   select {
      case <-time.After(delay):
      case <-p.ctx.Done():
         return
   }
...
```
If you don't see need for such test, I can remove it.

consensus_test.go
Stefan-Ethernal: > Most likely you can remove the go func() as well.

If I remove cancellation, state machine would transfer to `ValidateState` instead. I can change assertion easily, although I wanted specifically to test cancellation and make sure state machine is still in `AcceptState` (see https://github.com/0xPolygon/pbft-consensus/pull/15#discussion_r787730000).

consensus_test.go
Stefan-Ethernal: Purpose of this channel is to make sure that go routine started execution, before state machine began to run, in order to be able to cancel it immediately (while we are still in `AcceptState`). Any ideas on how to accomplish this behavior, but more elegant are welcome. :slightly_smiling_face:

consensus_test.go
```diff
@@ -208,6 +129,81 @@ func TestTransition_AcceptState_Validator_VerifyFails(t *testing.T) {
...
+		Data: mockProposal,
+		Time: time.Now().Add(time.Second),
+	})
+
+	go func() {
+		m.cancelFn()
+	}()
+
+	m.runCycle(m.ctx)
+	assert.True(t, m.IsState(AcceptState), "Expected to be in %s, but was in %s", AcceptState, m.getState())
```
ferranbt, line 204: Test is good, just remove this assertion. If we cancel the context from the outside, the state is not a result that we expect but just a side effect of the state machine not resetting it and using the default one. So, it is not an output of the cancellation.

consensus_test.go
```diff
@@ -208,6 +129,81 @@ func TestTransition_AcceptState_Validator_VerifyFails(t *testing.T) {
...
+		Type: MessageReq_Prepare,
+		View: ViewMsg(1, 0),
+	})
+	m.emitMsg(&MessageReq{
+		From: "C",
+		Type: MessageReq_Prepare,
+		View: ViewMsg(1, 0),
+	})
+
+	go func() {
```
ferranbt, line 162: What is happening here?

consensus_test.go
ferranbt: What about this?

ferranbt (changes requested):

consensus_test.go
Stefan-Ethernal: Ok. I'm just going to add assert.NotPanics, because it would be controversial for me to have a test without any assertions.

consensus_test.go
Stefan-Ethernal: I needed cancellation when in some previous iterations I had invoked Run function. If I haven't had cancellation, it would hang within the state machine loop until timeout at least or maybe even forever.
With runCycle function invocation, we are "safe", as it will execute only single cycle.

consensus_test.go
Stefan-Ethernal: Sorry, I have overseen this one.

consensus_test.go
ferranbt: This is resolved right? I cannot find it anymore.

ferranbt (approved):
Super nice PR!

consensus_test.go
Stefan-Ethernal: Channel remained in the code, I have just renamed it to `waitSignal`. The rationale I did it that way is described in previous comment (https://github.com/0xPolygon/pbft-consensus/pull/15#discussion_r787940759). I have tried to remove channel, but test becomes unstable and it depends when go routine begins executing.
Also the reason I haven't changed it to `runCycle`, but left `Run` function instead is because I wanted to cover main state machine loop as well in at least one test.

We can change test in some other PR if you want, but I am running out of ideas what to do here. 🤔

Stefan-Ethernal:
> Super nice PR!

The pleasure was all mine, nice and constructive review from your side.

PR #17, opened 2022-01-20
POS-130 - Fuzz tests

0xSasaPrsic:
Added fuzz tests for PBFT.

### DIFF
--- a/e2e/fuzz_unreliable_network_test.go
+++ b/e2e/fuzz_unreliable_network_test.go
@@ -1,0 +1,92 @@
+package e2e
+
+import (
+	"fmt"
+	"math/rand"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFuzz_Unreliable_Network(t *testing.T) {
+	rand.Seed(time.Now().Unix())
+	nodesCount := 20 + rand.Intn(11) // vary nodes [20,30]
+	maxFaulty := nodesCount/3 - 1
+	maxHeight := uint64(40)
+	currentHeight := uint64(0)
+	jitterMax := 500 * time.Millisecond
+	hook := newPartitionTransport(jitterMax)
+	c := newPBFTCluster(t, "network_unreliable", "prt", nodesCount, hook)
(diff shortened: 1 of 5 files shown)


e2e/fuzz_unreliable_network_test.go
```diff
@@ -0,0 +1,92 @@
...
+					if bp == v.name { // is part of the bigPartition
+						runningMajorityNodes = append(runningMajorityNodes, v.name)
+					}
+				}
+			} else {
+				stoppedNodes = append(stoppedNodes, v.name)
+			}
+		}
+		// check all running nodes in majority partition for the block height
+		fmt.Printf("Checking for height %v, started with nodes %d\n", currentHeight, nodesCount)
```
ferranbt, line 72: Use `t.Log` instead of Print.

e2e/fuzz_unreliable_network_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"math/rand"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFuzz_Unreliable_Network(t *testing.T) {
+	rand.Seed(time.Now().Unix())
```
ferranbt, line 14: Add a condition to only enabled this if `FUZZ` env variable is set.

ferranbt (changes requested):

e2e/fuzz_unreliable_network_test.go
0xSasaPrsic: Done.

e2e/fuzz_unreliable_network_test.go
ferranbt: Add this into a function
```
func isFuzzEnabled(t *testing.T) {
        if os.Getenv("FUZZ") != "true" {
		t.Skip("Fuzz tests are disabled.")
	}
}
```

ferranbt:
Could you enable fuzz tests in CI:
- Add a new command `make fuzz` that only runs the Fuzz tests (`--run Fuzz`)
- Add a parallel fuzz task in Github that only runs the fuzz tests with some 10m timeout.

ferranbt (approved):

PR #18, opened 2022-01-31
Add Init to Backend

ferranbt:
There is a likely scenario I am facing in which the `Backend` implementation stores locally some data about the current round in order to be reused between different stages of the Pbft cycle. For example, a Proposer can pre-compute all the state transition in the `Validate()` callback (`Accept` state) and only do a small canonical head change upon `Commit()`. This is not possible if we do not store some information on the Backend. However, that is not safe right now since a Round change might provoke that the data is not valid anymore.

This PR adds a new function to the `Backend` called `Init()`. This function is called every time that the `Accept` state is reached. During that function, the `Backend` should restart and flush any old data from its state from a previous round and this height (if any).

An alternative to this PR would be to do a `Reset()` function called whenever there is a round state change. However, this would not be called during the first iteration of the protocol and then it would be up to `BuildProposal()` and `Validate()` to run the same script to initialize the state. With `Init()` we get both things, a single entrypoint to initialize the state and to reset it.

### DIFF
--- a/consensus.go
+++ b/consensus.go
@@ -113,8 +113,8 @@ type Backend interface {
 	// Hash hashes the proposal bytes
 	Hash(p []byte) []byte
 
-	// Reset is used to signal the backend that the round is changing
-	Reset()
+	// Init is used to signal the backend that a new round is going to start.
+	Init()
 
 	// IsStuck returns whether the pbft is stucked
 	IsStuck(num uint64) (uint64, bool)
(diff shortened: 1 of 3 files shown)


0xSasaPrsic (approved):
LGTM!

consensus.go
```diff
@@ -113,6 +113,9 @@ type Backend interface {
 	// Hash hashes the proposal bytes
 	Hash(p []byte) []byte
 
+	// Init is used to signal the backend that a new round is going to start.
+	Init()
```
stefan-negovanovic, line 117: Have you considered maybe that this function returns an `error` object or `bool` in case it fails? Not sure how trivial logic within it would be and is there any scenario where `Init` function fails and it is needed to be handled somehow from the "outside"?

stefan-negovanovic (commented):
LGTM

Stefan-Ethernal (commented):
.

PR #19, opened 2022-02-01
POS-58 IBFT liveness issues

Stefan-Ethernal:
This PR implements integration test that proves that PolyBFT consensus algorithm (inspired by IBFT) has liveness issues. Test is currently deliberately failing, because it expects that there are no errors, however timeout error is triggered by test framework since no consensus is reached after predefined time.

Test is implemented by having a scenario, discussed in [Correctness Analysis of Istanbul Byzantine Fault Tolerance paper, Chapter 7.1, Case 1](https://arxiv.org/pdf/1901.07160.pdf) on mind:
Note: for simplicity, imagine that we have validator set which has 5 validators {a1, a2, a3, a4 and a5}. Validator nodes are not named exactly this way within the test, but the point is there.
- a3 is proposer and proposes block B (sends PREPARE and PRE-PREPARE messages).
- a1 and a2 receive PRE-PREPARE from all the validator nodes and send PREPARE messages. a3, a4 and a5 don't receive either message (e.g. transient network issues).
- a1 and a2 locked on proposal B, since they have received enough PREPARE messages (2*F+1, where F is number of max faulty nodes),
- since there are only PREPARE messages from a1 and a2 and no COMMIT messages, when timeout elapses, all the nodes move to ROUND CHANGE state,
- when they eventually gossip enough ROUND CHANGE messages, they agree upon moving to the next round where a new proposer (e.g. a4) gets calculated,
- a4 proposes new proposal B' which is different from the one a1 and a2 got locked on,
- a3, a4 and a5 all respond with PREPARE message and lock on B' proposal (whereas a1 and a2 respond with ROUND CHANGE, since they are already locked on proposal B),
- before responding with COMMIT messages, some of those nodes becomes faulty (e.g. a3) and only a4 and a5 are able to send COMMIT messages. So we don't have enough COMMIT messages in order to move to COMMIT state (and eventually make a final proposal). 

We have situation where a1 and a2 make partition 1 (locked on proposal B) and a3 and a5 (locked on proposal B'). The faulty a4 node is lacking in order to have quorum.


zivkovicmilos:
@Stefan-Ethernal 

Why did you remove the liveness simulation test in commit 53103f7?

Stefan-Ethernal:
> @Stefan-Ethernal
> 
> Why did you remove the liveness simulation test in commit [a4de07b](https://github.com/0xPolygon/pbft-consensus/commit/a4de07b1fa0100bfe9508bcad3c8aa4d8bb9e062)?

Hey @zivkovicmilos, we have removed it, because we are finding that test too artificial (since most of behavior is mocked and we have full control over PBFT message injecting) and I'm afraid that it doesn't truly proves existence of liveness issue. That's why we have implemented E2E test, which seems that proves existence of liveness issue in a more real use case manner.
However, liveness simulation test emulates steps which are leading to the liveness issue and maybe it is useful to some extent to see more precisely which steps are leading to this issue... We can discuss whether to eventually keep it, if you see some value in it.

zivkovicmilos:
@Stefan-Ethernal 

I'm assuming you're talking about [this e2e test](https://github.com/0xPolygon/pbft-consensus/blob/53103f7490f491361be173af9f3d3e11bc788207/e2e/e2e_partition_test.go#L13-L101)?

Stefan-Ethernal:
> @Stefan-Ethernal
> 
> I'm assuming you're talking about [this e2e test](https://github.com/0xPolygon/pbft-consensus/blob/53103f7490f491361be173af9f3d3e11bc788207/e2e/e2e_partition_test.go#L13-L101)?

Yes that's the one. :slightly_smiling_face:

zivkovicmilos:
@Stefan-Ethernal 

I don't understand 1 thing though - where do you actually simulate the steps laid out in the liveness analysis?
I can't find the place where you explicitly check locking, where you explicitly send out Preprepare messages, and shut down the Byzantine node.

You're supposed to simulate 2 subsections of nodes, 1 that's locked onto block `B`, and another that's locked onto block `B'`. The second subsection of nodes should lock onto `B'` after all the nodes in the subsection broadcast their preprepare messages (after a round change in the `B` subsection) and one Byzantine node goes offline - that's how you ensure that there is no block production in any subsection, since each subsection doesn't have enough nodes to reach consensus and they're both locked onto blocks `B` and `B'`, respectively.

Can you please point out the sections of this liveness test that cover these steps?

Stefan-Ethernal:
@zivkovicmilos 

> I don't understand 1 thing though - where do you actually simulate the steps laid out in the liveness analysis?

We are simulating those steps on transport level (routing messages selectively to targeted nodes) and that's how we assure that only subset of nodes receive certain messages. The key part of the `TestE2E_Partition_Liveness` is `livenessGossipArbitrage` callback, which has all the logic for the messages routing. You can see that we have some "routing table" for two rounds (which is enough to simulate required steps). Consequently, subsets of nodes lock on different proposals.

> I can't find the place where you explicitly check locking,...

We aren't explicitly checking locking and proposals (although we could add it), but we have added logging, which clearly shows that we have two nodes partitions locked on different proposals. 

> ...where you explicitly send out Preprepare messages,...

Preprepare, as well as other messages are sent from Pbft state machine itself, so there is no need to send it explicitly (as we have needed to do with `Test_LivenessIssueSimulation` unit test). We just need to route messages, namely cut out any undesired communication within gossip process.

> ...and shut down the Byzantine node.

There is no need to shut down Byzantine node, it is enough to isolate it from communication (e.g. unable to send or receive messages) and therefore in our example of 5 nodes in total, second partition lacks one validator vote (the one we have proclaimed as faulty) in order to promote their locked proposal as a final one. Faulty node is simulated with this if statement in `livenessGossipArbitrage`: `if sender == faultyNodeId || receiver == faultyNodeId`

So just to summarize: the whole point of the test was to let Pbft machinery run on its own (which is near real-life scenario), although we need control over message routing (in order to make sure that only targeted nodes receive certain messages and to induce they lock on different proposals). We let the test run long enough (currently it is set to 5 minutes, or approximately 8 rounds) and checked against timeout error, which appears from the test framework, since there is no progress regarding proposal finalization.

ferranbt:
I agree with @Stefan-Ethernal. The unit tests look nice right now because they test something simple, this is, simple state transitions in the pbft node. Once you test more difficult stuff, the complexity grows really fast. So the general idea for the tests is: unit test for single node behavior (can the node move from state A to B?) and integration tests for multi node behavior (can the noed recover from state A when the others are in state B and C?).

consensus.go
```diff
@@ -181,6 +181,18 @@ func New(validator SignKey, transport Transport, opts ...ConfigOption) *Pbft {
...
 }
 
+// IsStateLocked only for debug purpose
+// TODO: remove this
+func (p *Pbft) IsStateLocked() bool {
+	return p.state.IsLocked()
+}
+
+// Proposal only for debug purpose
+// TODO: remove this
```
ferranbt, line 191: Do we still need this?

e2e/e2e_partition_test.go
```diff
@@ -6,9 +6,100 @@ import (
 	"testing"
 	"time"
 
+	"github.com/0xPolygon/pbft-consensus"
 	"github.com/stretchr/testify/assert"
 )
 
+func TestE2E_Partition_Liveness(t *testing.T) {
```
ferranbt, line 13: Please add a command here with a reference to the paper so that we know exactly what issue is this covering.

ferranbt (changes requested):

consensus.go
Stefan-Ethernal: Currently at the end of the test, we are logging locked indication and actual proposal for all the nodes, which is convenient to verify everything is as expected:
`t.Logf("Node %v, isProposalLocked: %v, proposal data: %v\n", n.name, n.pbft.IsStateLocked(), n.pbft.Proposal().Data)`

However, we are thinking through some of the options and not sure what is the best:
1. keep this log and just remove those TODO comments for these two methods,
2. whether it would be better to turn those logs into asserts, where we should check that proposals are different and that required nodes are locked,
3. remove that log within the test, keep `WaitForHeight` and asserting that there is no (timeout) error after invoking it. In that case, we can remove those two methods.

WDYT is the best solution or can you come up something different, more convenient?

ferranbt: For now, I think I am fine with 1 to be honest.

zivkovicmilos:
@Stefan-Ethernal 

Thank you for the detailed explanation 🙏 

There are a couple of things I'd like to point out here, which I think are needed in order for this e2e test to be valid:
* There is no need to explicitly wait 5min, or any time for that matter in any kind of test - it's usually an indication of a bad test design and should be refactored to rely on a more precise metric
* Why are you relying on logs to make this test work? Do you plan to remove them at some point? You should add explicit locking and state asserts in the test itself
* The test itself lacks documentation, even though the explanation you've provided to me beforehand is sufficient - but it's not present in either the PR description or finely presented in code

cc @ferranbt

Stefan-Ethernal:
@zivkovicmilos
> There is no need to explicitly wait 5min, or any time for that matter in any kind of test - it's usually an indication of a bad test design and should be refactored to rely on a more precise metric.
EDIT: I see that you're waiting on the height to be reached in a happy path scneario, but if you know the height will never be reached then timeouts shouldn't be used at all

I'm not sure whether it is correct or not, but **maybe** we could track round numbers and wait for predefined number of rounds. Anyhow, since all the other tests are using this way to check whether there is a proposal or not, we thought it is correct to use it here as well. Bear in mind that this is not a unit test, but end to end test, therefore we have pretty limited maneuvering space when it comes to checking about proposal finalization.

> Why are you relying on logs to make this test work? Do you plan to remove them at some point? You should add explicit locking and state asserts in the test itself

We don't need logs in order to test work (waiting for block height for predefined amount of time is enough to prove existence of liveness issue). Although agree that maybe we could add assertions to make sure that nodes are locked on different proposals (which is some of possibilities mentioned to @ferranbt in https://github.com/0xPolygon/pbft-consensus/pull/19#discussion_r797402001, but based on his reply seems that we are good to go with this solution, at least for now).

> The test itself lacks documentation, even though the explanation you've provided to me beforehand is sufficient - but it's not present in either the PR description or finely presented in code

We have added comment on test in a mean time and described what test simulates. If you want review it and see whether we can make it even better and more clear.

ferranbt:
> There is no need to explicitly wait 5min, or any time for that matter in any kind of test - it's usually an indication of a bad test design and should be refactored to rely on a more precise metric.

These integration tests (not unit) are the same to what it is being done right now on [edge](https://github.com/0xPolygon/polygon-edge/blob/43fb2a6805711919bdb828d6e7873f01eb2ff2b2/e2e/framework/helper.go#L472) to test a similar situation, you just run a cluster of machines and expect some result under certain timeout conditions. The only difference is that we wait here for 5 minutes instead of 1.

Note that we are only waiting that much because the test fails. Once the fix is done in a followup PR, the test will pass and we will not wait for 5 min.

Please @Stefan-Ethernal update this PR and merge into a "liveness-issue" branch, we will do the followup PRs on that branch as well.

zivkovicmilos:
@ferranbt 
>These integration tests (not unit) are the same to what it is being done right now on edge to test a similar situation, you just run a cluster of machines and expect some result under certain timeout conditions. The only difference is that we wait here for 5 minutes instead of 1.

Just because we use it on a separate repo doesn't make it any more correct - we are currently refactoring the entire testing suite on Edge to weed out any timeouts, as they're in their essence faulty and not in any way a metric of a successful test execution.

I'm just saying this e2e test assumes way too much on the fly, and doesn't check everything that it needs to check. The bulk of the logic in `livenessGossipArbitrage` is convoluted and not very clear without an explanation to accompany it.
Not to mention it prints out stuff explicitly to the logs during execution.

The test only covers a single case, and doesn't cover / test all of the other cases mentioned in the analysis doc. 
I'll leave a review on my 2 cents regarding this PR 👍

consensus.go
```diff
@@ -181,6 +181,14 @@ func New(validator SignKey, transport Transport, opts ...ConfigOption) *Pbft {
 	return p
 }
 
+func (p *Pbft) IsStateLocked() bool {
+	return p.state.IsLocked()
+}
+
+func (p *Pbft) Proposal() *Proposal {
```
zivkovicmilos, line 188: Method should be called `GetProposal`

e2e/e2e_partition_test.go
```diff
@@ -6,9 +6,110 @@ import (
...
+// In the round=1, it marks one node as faulty (meaning that it doesn't takes part in gossiping).
+func TestE2E_Partition_Liveness(t *testing.T) {
+	const nodesCnt = 5
+	round0 := roundMetadata{
+		round: 0,
+		// lock A_3 and A_4 on one proposal
+		routingMap: map[pbft.NodeID][]pbft.NodeID{
+			"A_0": {"A_3", "A_4"},
+			"A_3": {"A_0", "A_3", "A_4"},
+			"A_4": {"A_3", "A_4"},
```
zivkovicmilos, line 31: What is specifically mapped here?

Additional in-code explanations would be welcome 👍

e2e/e2e_partition_test.go
```diff
@@ -6,9 +6,110 @@ import (
...
+	round1 := roundMetadata{
+		round: 1,
+		// lock A_2 and A_0 on one proposal and A_1 will be faulty
+		routingMap: map[pbft.NodeID][]pbft.NodeID{
+			"A_0": {"A_0", "A_2", "A_3", "A_4"},
+			"A_1": {"A_0", "A_2", "A_3", "A_4"},
+			"A_2": {"A_0", "A_1", "A_2", "A_3", "A_4"},
+
+			"A_3": {"A_0", "A_1", "A_2", "A_3", "A_4"},
+			"A_4": {"A_0", "A_1", "A_2", "A_3", "A_4"},
```
zivkovicmilos, line 44: Same as in here:
https://github.com/0xPolygon/pbft-consensus/pull/19/files#r797674810

e2e/e2e_partition_test.go
```diff
@@ -6,9 +6,110 @@ import (
...
+	livenessGossipArbitrage := func(sender, receiver pbft.NodeID, msg *pbft.MessageReq) bool {
+		faultyNodeId := pbft.NodeID("A_1")
+		if msg.View.Sequence > 2 || msg.View.Round > 1 {
+			// node A_1 (faulty) is unresponsive after round 1
+			if sender == faultyNodeId || receiver == faultyNodeId {
+				return false
+			}
+			// all other nodes are connected for all the messages
+			return true
+		}
```
zivkovicmilos, line 58: This should be a separate method

e2e/e2e_partition_test.go
```diff
@@ -6,9 +6,110 @@ import (
...
+
+		// Case where we are in round 1 and 2 different nodes will lock the proposal
+		// (A_1 ignores round change and commit messages)
+		if msg.View.Round == 1 {
+			if sender == faultyNodeId &&
+				(msg.Type == pbft.MessageReq_RoundChange || msg.Type == pbft.MessageReq_Commit) {
+				return false
+			}
+		}
```
zivkovicmilos, line 67: This should be a separate method

e2e/e2e_partition_test.go
```diff
@@ -6,9 +6,110 @@ import (
...
+
+			foundReceiver := false
+			for _, v := range receivers {
+				if v == receiver {
+					foundReceiver = true
+					break
+				}
+			}
+			return foundReceiver
+		}
```
zivkovicmilos, line 93: This should be a separate method, simplified.

Inside the block you rely on checks that happened 30 lines ago

e2e/e2e_partition_test.go
```diff
@@ -6,9 +6,110 @@ import (
...
+	c := newPBFTCluster(t, "liveness_issue", "A", nodesCnt, hook)
+	c.Start()
+	defer c.Stop()
+
+	err := c.WaitForHeight(3, 5*time.Minute)
+
+	// log to check what is the end state
+	for _, n := range c.nodes {
+		t.Logf("Node %v, isProposalLocked: %v, proposal data: %v\n", n.name, n.pbft.IsStateLocked(), n.pbft.Proposal().Data)
+	}
```
zivkovicmilos, line 108: This log shouldn't belong in the test

e2e/framework.go
```diff
@@ -422,8 +423,11 @@ func (f *fsm) IsStuck(num uint64) (uint64, bool) {
 }
 
 func (f *fsm) BuildProposal() (*pbft.Proposal, error) {
+	// make different proposal for each sequence/round
+	prop := make([]byte, 8)
+	_, _ = rand.Read(prop)
```
zivkovicmilos, line 428: You're using a weak number generator here (nitpick), but also not checking the errors it gives you

e2e/framework.go
```diff
@@ -422,8 +423,11 @@ func (f *fsm) IsStuck(num uint64) (uint64, bool) {
 }
 
 func (f *fsm) BuildProposal() (*pbft.Proposal, error) {
+	// make different proposal for each sequence/round
+	prop := make([]byte, 8)
```
zivkovicmilos, line 427: Please extract these magic numbers to a constant (`8`)

e2e/transport.go
```diff
@@ -34,6 +34,7 @@ func (t *transport) Gossip(msg *pbft.MessageReq) error {
 				send = t.hook.Gossip(msg.From, to, msg)
 			}
 			if send {
+				// fmt.Printf("Sending message %v from Node %v to node %v\n", msg.Type, msg.From, to)
```
zivkovicmilos, line 37: Leftover print

e2e/transport.go
```diff
@@ -46,6 +47,37 @@ type transportHook interface {
 	Gossip(from, to pbft.NodeID, msg *pbft.MessageReq) bool
 }
 
+// encapsulates routing mapping for certain round
+type roundMetadata struct {
+	round uint64
+	// represents message routing map (sender node id to recipient node ids)
+	// e.g. A4 sends messages to [A0, A1]
```
zivkovicmilos, line 54: If you need to explain the usage of the field in a comment, it's probably not named right

e2e/transport.go
```diff
@@ -46,6 +47,37 @@ type transportHook interface {
...
+	routingMap map[pbft.NodeID][]pbft.NodeID
+}
+
+// callback which enables determining which message should be gossiped
+type gossipArbitrageHandler func(sender, reciever pbft.NodeID, msg *pbft.MessageReq) (shouldSend bool)
```
zivkovicmilos, line 59: Why does the type contain the structure `arbitrage`?

e2e/transport.go
```diff
@@ -46,6 +47,37 @@ type transportHook interface {
...
+}
+
+func newGenericGossipTransport(gossipArbitrage gossipArbitrageHandler) *genericGossipTransport {
+	return &genericGossipTransport{
+		gossipArbitrage: gossipArbitrage,
+	}
+}
+
+func (rt *genericGossipTransport) Gossip(from, to pbft.NodeID, msg *pbft.MessageReq) bool {
+	// if gossip arbitrage callback returns false, message should not be gossiped and we stop here.
```
zivkovicmilos, line 73: This should be part of the method definition comment, not an implementation comment

e2e/transport.go
```diff
@@ -46,6 +47,37 @@ type transportHook interface {
...
+	return rt.gossipArbitrage(from, to, msg)
+}
+
+func (rt *genericGossipTransport) Connects(from, to pbft.NodeID) bool {
```
zivkovicmilos, line 77: Method name should be probably `ShouldConnectTo`, since `Connects` is to vague, and this is a public method

e2e/transport.go
zivkovicmilos, line 59: Typo here `reciever` should be `receiver`

zivkovicmilos (changes requested):
I've left some comments on the PR 👍 

Overall, I really don't like the idea of using a clunky map in order to facilitate message gossiping and round flows - this is the major part that I would like to see refactored

e2e/e2e_partition_test.go
```diff
@@ -6,9 +6,110 @@ import (
...
+// Correctness Analysis of Istanbul Byzantine Fault Tolerance(https://arxiv.org/pdf/1901.07160.pdf).
+// Specific problem this test is emulating is described in Chapter 7.1, Case 1.
+// Summary of the problem is that there are not enough nodes to lock on single proposal,
+// due to some issues where nodes being unable to deliver messages to all of the peers.
+// Therefore nodes are split into two subsets locked on different proposals.
+// Simulating one node from larger subset as faulty one, results in being unable to reach a consensus on a single proposal and that's what liveness issue is about.
+// This test creates a single cluster of 5 nodes, but instead of letting all the peers communicate with each other,
+// it routes messages only to specific nodes and induces that nodes lock on different proposal.
+// In the round=1, it marks one node as faulty (meaning that it doesn't takes part in gossiping).
+func TestE2E_Partition_Liveness(t *testing.T) {
```
zivkovicmilos, line 23: This test only covers a specific single case - it doesn't cover the other edge cases the analysis doc mentions.

Also, some assertions are missing from the test, as you assume they are correctly set in the background

Stefan-Ethernal:
> I've left some comments on the PR +1

Thanks for taking your time to review it. :+1: We will analyze and try to address as many things as possible.

> Overall, I really don't like the idea of using a clunky map in order to facilitate message gossiping and round flows - this is the major part that I would like to see refactored

Any suggestions how to refactor that part? We didn't want to introduce additional complexity and over-engineer a test, so this seemed like good enough approach.

e2e/transport.go
Stefan-Ethernal: First slice of implementation implied that only specific part for this test is contained within the delegate and if delegate doesn't manage to handle it, in that case messages are routed based on provided `routingMap` within the `Gossip` function itself. However this is now obsolete, since all the gossip logic is contained within the callback, therefore you are right, we will remove arbitrage word from it, so it will be called `gossipHandler`.

e2e/e2e_partition_test.go
Stefan-Ethernal: Will make this part shorter, so there is no need for separate function.

e2e/e2e_partition_test.go
Stefan-Ethernal: Similar goes as with https://github.com/0xPolygon/pbft-consensus/pull/19#discussion_r798530551.

e2e/e2e_partition_test.go
Stefan-Ethernal: Agreed, this part is going to be extracted out from this body and a function will be extracted:

```
// do not send commit messages for rounds <=1
if msg.Type == pbft.MessageReq_Commit {
    return false
}
```

e2e/framework.go
Stefan-Ethernal: > You're using a weak number generator here (nitpick - you can use crypto/rand)

It is fine for our use case, since we need different proposal from each round.

> but also not checking the errors it gives you

It is never returning an error, it is there as a return value because of interface. 
From the documentation: "Read generates len(p) random bytes from the default Source and writes them into p. **It always returns len(p) and a nil error**."

Stefan-Ethernal:
Due to branch renames, this PR closed automatically. Therefore I have created another one (PR https://github.com/0xPolygon/pbft-consensus/pull/22) as a follow up. It has all the changes as before+fixes with findings from @zivkovicmilos.

PR #20, opened 2022-02-02
Add String() to View

ferranbt:
When I print the `View` struct with the `fmt` library due to the `Round` field being before the `Sequence` field it shows as (Round, Sequence) on the screen which is confusing. Instead of changing the order of the fields, this PR implements the `String()` method directly.

### DIFF
--- a/state.go
+++ b/state.go
@@ -84,7 +84,7 @@ func (v *View) Copy() *View {
 }
 
 func (v *View) String() string {
-	return fmt.Sprintf("(%d, %d)", v.Sequence, v.Round)
+	return fmt.Sprintf("(Sequence=%d, Round=%d)", v.Sequence, v.Round)
 }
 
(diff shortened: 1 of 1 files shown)


state.go
```diff
@@ -83,6 +83,10 @@ func (v *View) Copy() *View {
 	return vv
 }
 
+func (v *View) String() string {
+	return fmt.Sprintf("(%d, %d)", v.Sequence, v.Round)
```
zivkovicmilos, line 87: Do you think it would be wise to add additional information to this print?
How would people know if a round or sequence come first, apart from diving into the code?

What do you think about making the print something like:
```go
return fmt.Sprintf("(Sequence [%d]; Round [%d])", v.Sequence, v.Round)
```

0xSasaPrsic (approved):
LGTM!

PR #22, opened 2022-02-03
POS-58 Implement liveness issue test proof

Stefan-Ethernal:
PR is follow up of stale PR https://github.com/0xPolygon/pbft-consensus/pull/19

This PR implements integration test that proves that PolyBFT consensus algorithm (inspired by IBFT) has liveness issues. Test is currently deliberately failing, because it expects that there are no errors, however timeout error is triggered by test framework since no consensus is reached after predefined time.

Test is implemented by having a scenario, discussed in [Correctness Analysis of Istanbul Byzantine Fault Tolerance paper, Chapter 7.1, Case 1](https://arxiv.org/pdf/1901.07160.pdf) on mind:
Note: for simplicity, imagine that we have validator set which has 5 validators {a1, a2, a3, a4 and a5}. Validator nodes are not named exactly this way within the test, but the point is there.

- a3 is proposer and proposes block B (sends PREPARE and PRE-PREPARE messages),
- a1 and a2 receive PRE-PREPARE from all the validator nodes and send PREPARE messages. a3, a4 and a5 don't receive either message (e.g. transient network issues),
- a1 and a2 locked on proposal B, since they have received enough PREPARE messages (2*F+1, where F is number of max faulty nodes),
- since there are only PREPARE messages from a1 and a2 and no COMMIT messages, when timeout elapses, all the nodes move to ROUND CHANGE state,
- when they eventually gossip enough ROUND CHANGE messages, they agree upon moving to the next round where a new proposer (e.g. a4) gets calculated,
- a4 proposes new proposal B' which is different from the one a1 and a2 got locked on,
- a3, a4 and a5 all respond with PREPARE message and lock on B' proposal (whereas a1 and a2 respond with ROUND CHANGE, since they are already locked on proposal B),
- before responding with COMMIT messages, some of those nodes becomes faulty (e.g. a3) and only a4 and a5 are able to send COMMIT messages. So we don't have enough COMMIT messages in order to move to COMMIT state (and eventually make a final proposal).

We have situation where a1 and a2 make partition 1 (locked on proposal B) and a3 and a5 (locked on proposal B'). The faulty a4 node is lacking in order to have quorum.

### DIFF
--- a/e2e/e2e_partition_test.go
+++ b/e2e/e2e_partition_test.go
@@ -6,8 +6,109 @@ import (
 	"testing"
 	"time"
 
+	"github.com/0xPolygon/pbft-consensus"
 	"github.com/stretchr/testify/assert"
 )
+
+// Test proves existence of liveness issues which is described in
+// Correctness Analysis of Istanbul Byzantine Fault Tolerance(https://arxiv.org/pdf/1901.07160.pdf).
+// Specific problem this test is emulating is described in Chapter 7.1, Case 1.
+// Summary of the problem is that there are not enough nodes to lock on single proposal,
+// due to some issues where nodes being unable to deliver messages to all of the peers.
+// Therefore nodes are split into two subsets locked on different proposals.
+// Simulating one node from larger subset as faulty one, results in being unable to reach a consensus on a single proposal and that's what liveness issue is about.
+// This test creates a single cluster of 5 nodes, but instead of letting all the peers communicate with each other,
+// it routes messages only to specific nodes and induces that nodes lock on different proposal.
+// In the round=1, it marks one node as faulty (meaning that it doesn't takes part in gossiping).
+func TestE2E_Partition_Liveness(t *testing.T) {
+	const nodesCnt = 5
+	round0 := roundMetadata{
+		round: 0,
+		// lock A_3 and A_4 on one proposal
+		routingMap: map[pbft.NodeID][]pbft.NodeID{
+			"A_0": {"A_3", "A_4"},
+			"A_3": {"A_0", "A_3", "A_4"},
+			"A_4": {"A_3", "A_4"},
+		},
+	}
+
+	round1 := roundMetadata{
+		round: 1,
+		// lock A_2 and A_0 on one proposal and A_1 will be faulty
+		routingMap: map[pbft.NodeID][]pbft.NodeID{
+			"A_0": {"A_0", "A_2", "A_3", "A_4"},
+			"A_1": {"A_0", "A_2", "A_3", "A_4"},
+			"A_2": {"A_0", "A_1", "A_2", "A_3", "A_4"},
+
+			"A_3": {"A_0", "A_1", "A_2", "A_3", "A_4"},
+			"A_4": {"A_0", "A_1", "A_2", "A_3", "A_4"},
+		},
+	}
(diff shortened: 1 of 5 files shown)


e2e/framework.go
```diff
@@ -422,17 +425,16 @@ func (f *fsm) IsStuck(num uint64) (uint64, bool) {
 }
 
 func (f *fsm) BuildProposal() (*pbft.Proposal, error) {
+	// make different proposal for each sequence/round
+	prop := make([]byte, ProposalLength)
```
ferranbt, line 429: Please revert this change. This is just a test, we do not need to do this here.

e2e/transport.go
```diff
@@ -42,10 +42,41 @@ func (t *transport) Gossip(msg *pbft.MessageReq) error {
 }
 
 type transportHook interface {
-	Connects(from, to pbft.NodeID) bool
```
ferranbt, line 45: Please revert this change. I am fine changing the naming of the interface but not as part of a PR that is addressing something else.

e2e/e2e_partition_test.go
```diff
@@ -1,14 +1,133 @@
...
+	hook := newGenericGossipTransport(livenessGossipHandler)
+
+	c := newPBFTCluster(t, "liveness_issue", "A", nodesCnt, hook)
+	c.Start()
+	defer c.Stop()
+
+	err := c.WaitForHeight(3, 5*time.Minute)
+
+	// Query nodes and make appropriate assertions
+	nodeSubsetA := []pbft.NodeID{"A_0", "A_1", "A_2"}
```
ferranbt, line 81: Please revert this change as well. We do not need to do an extra check to see if there are two locked proposals, it is enough for me if we log it. This is the idea behind the integration tests, to test some specific multi node situation and wait for the chain to finish. I do not want to over complicate it this way.

e2e/e2e_partition_test.go
```diff
@@ -1,14 +1,133 @@
...
+	nodeA0 := c.nodes[string(nodeSubsetA[0])]
+	nodeA3 := c.nodes[string(nodeSubsetB[0])]
+	queryNodesFromSubset(t, c.nodes, faultyNodeId, nodeA0, nodeSubsetA)
+	queryNodesFromSubset(t, c.nodes, faultyNodeId, nodeA3, nodeSubsetB)
+
+	assert.NoError(t, err)
+}
+
+// Helper function determining whether a message should be gossiped, based on round message flow.
+func shouldGossipBasedOnMsgFlowMap(flowMap map[uint64]roundMetadata, msg *pbft.MessageReq, senderId pbft.NodeID, receiverId pbft.NodeID) bool {
```
ferranbt, line 92: Could you add this function as a method of `roundMetadata`?
```
func (r *roundMetadata) shouldGossipBasedOnMsgFlowMap...
```

ferranbt (changes requested):

e2e/e2e_partition_test.go
Stefan-Ethernal: Seems like it more suits to the `genericGossipTransport`, so I have moved it there. Hope that you don't mind.

ferranbt (approved):
LGTM! Lets merge this and work on the fix.

PR #23, opened 2022-02-10
Decouple MaxFaultyNodes and QuorumSize calculations from currentState struct

Stefan-Ethernal:
#### Max faulty nodes calculation
When working on fuzz framework, it was realized that `MaxFaultyNodes` calculation is needed there as well, so this PR decouples the calculation from the `currentState` struct, by making it available on the package-level in order to be able to reuse it, without code duplication.

#### Quorum size calculation
As mentioned in comment https://github.com/0xPolygon/pbft-consensus/pull/23#issuecomment-1034972046, additional package-level function is introduced called `QuorumSize`. It determines minimum number of messages needed upon agreement when moving from one to another `PbftState` within the state machine.

### DIFF
--- a/consensus.go
+++ b/consensus.go
@@ -762,3 +762,21 @@ func exponentialTimeout(round uint64) time.Duration {
 	}
 	return timeout
 }
+
+// Calculate max faulty nodes in order to have Byzantine-fault tollerant system.
+// Formula explanation:
+// N -> number of nodes in PBFT
+// F -> number of faulty nodes
+// N = 3 * F + 1 => F = (N - 1) / 3
+
+// PBFT tolerates 1 failure with 4 nodes
+// 4 = 3 * 1 + 1
+// To tolerate 2 failures, PBFT requires 7 nodes
+// 7 = 3 * 2 + 1
+// It should always take the floor of the result
+func MaxFaultyNodes(nodesCount int) int {
+	if nodesCount <= 0 {
+		return 0
+	}
+	return (nodesCount - 1) / 3
+}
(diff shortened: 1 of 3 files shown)


Stefan-Ethernal:
Also not sure about `NumValid` function, since my gut feeling is that it should be also be available on a package level as helper function. Consider renaming it to `QuorumSize` and use it instead of `NumValid`, just carefully adapt logical conditions (since `NumValid` implicitly implies +1 part in 2*F+1 formula).

consensus.go
```diff
@@ -762,3 +762,21 @@ func exponentialTimeout(round uint64) time.Duration {
...
+// N -> number of nodes in PBFT
+// F -> number of faulty nodes
+// N = 3 * F + 1 => F = (N - 1) / 3
+
+// PBFT tolerates 1 failure with 4 nodes
+// 4 = 3 * 1 + 1
+// To tolerate 2 failures, PBFT requires 7 nodes
+// 7 = 3 * 2 + 1
+// It should always take the floor of the result
+func MaxFaultyNodes(nodesCount int) int {
```
ferranbt, line 777: Can you add unit tests for this?

ferranbt (changes requested):
Small nitpick here. Where do you expect to use `NumValid`? I am good moving it to a helper function as well.

ferranbt:
That is a nice idea. I am okay with that.

Stefan-Ethernal:
> That is a nice idea. I am okay with that.

Ok, in that case I will convert PR to draft and let you know when have implemented QuorumSize as well.

consensus.go
Stefan-Ethernal: There already was a unit test for this part, just added `TotalNodesCount=0` in order to cover that sanity check logical branch within the calculation as well.

ferranbt (approved):

PR #24, opened 2022-02-10
Fuzz framework support

0xSasaPrsic:
### Fuzz daemon
Fuzz daemon represents a runner, implemented as a go routine. It initializes and run a cluster of predefined nodes count and apply different actions to it (currently only a single, drop node action). Intention is that actions are randomly picked and grouped into scenarios, but that part will be addressed through follow up PRs.
### Drop node action
This PR implements periodically dropping predefined count of nodes and runner currently executes only that action.
### Next steps
- [ ] ~~Finish scenario support implementation and randomize actions picking and grouping it into scenarios~~. Obsolete requirement.
- [x] Introduce FunctionalAction interface, which should have `CanApply` and `Apply` functions. If `CanApply` returns `false`, the action is skipped. Apply returns `RevertFunc` callback, which is called once action reversal takes place.
- [x] Simplify `DropNodeAction`, ~~by extracting periodical logic into the separate action (called `PeriodicAction`), which should receive action as a parameter and execute it periodically (N times in predefined time intervals).~~ `Runner` is in charge of periodical actions execution (applying and reverting).
- [ ] Introduce complex hooks support (single transport hook should be as simple as possible and composing hooks by adding it into some array within the cluster)

Additional actions implementation:
- [x] `AddLatencyAction`
- [x] `CreatePartitionsAction`
- [x] `FlowMapAction` (?)


e2e/actions.go
```diff
@@ -0,0 +1,120 @@
...
+type Action interface {
+	Apply(c *Cluster)
+	Revert(c *Cluster)
+}
+
+// This action encapsulates logic for dropping nodes action.
+type DropNodeAction struct {
+	dropCount                int
+	dropProbabilityThreshold int
+	// TODO: Idea: Extract these two fields to another action (called PeriodicAction), which would receive as a parameter another action and would run it?
```
ferranbt, line 40: I like the idea of modularity and smart design but lets make sure that going into this road of `Actions` inside `Actions` does not become a problem in the end. As far as I can think of, we only have three attack vectors? Drop node, latency and partition (flow map here). I can't think of anything else.

e2e/actions.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (dn *DropNodeAction) Apply(c *Cluster) {
+	if err := dn.Validate(c); err != nil {
+		log.Printf("[WARNING] Skipping drop node action. Reason: '%s'", err)
+		return
+	}
+
+	maxFaultyNodes := GetMaxFaultyNodes(len(c.nodes))
+	ticker := time.NewTicker(dn.droppingInterval)
+	after := time.After(dn.duration)
```
ferranbt, line 78: I would prefer to have the actions as more functional objects. For example, `Apply()` only checks if its possible to drop a node and drops it. It might return a callback to revert this specific drop. Then, it is up to a higher level function like the `runner` to apply this loop that applies and reverts actions.

Otherwise, as it is right now, you are going to start a lot of concurrent operations that might produce the type of lock issues you have been experiencing. Specially, when the behaviour of executing an action with a certain probability can be easily extracted to the runner. The actions are simplified even more.

ferranbt (changes requested):

e2e/actions.go
Stefan-Ethernal: The goal of creating `PeriodicAction` was just to simplify `DropNodeAction`, because `droppingInterval` and `duration` doesn't seem like they belong to the `DropNodeAction` (which is intended to be as simple as possible). Periodical execution of actions should be handled somehow from the outside (either by encapsulating that logic into the `PeriodicAction` struct or to let `Runner` take care of that etc.)

e2e/actions.go
Stefan-Ethernal: As I have stated in the https://github.com/0xPolygon/pbft-consensus/pull/24#discussion_r805644418, I agree that periodical logic execution should be removed from `DropNodeAction` and let some higher-level abstraction take care of that part instead.
Regarding probabilistic execution of actions, just want to make sure that I have understood correctly: `Runner` should execute randomly picked `Action` with certain probability (`DropNodeAction` should not take care of that part either). Is that correct?

However is it ok to keep `Action` struct, since it seems as a cleaner solution to me, than having a bunch of functions with a callback revert function as a parameter? Maybe you can take a look at the PR https://github.com/0xPolygon/pbft-consensus/pull/25, since it depends on this one, if we are going to change the way we are representing actions.

e2e/actions.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"fmt"
+	"log"
+	"math/rand"
+
+	"github.com/0xPolygon/pbft-consensus"
+)
+
+type Action interface {
+	Apply(c *Cluster)
```
ferranbt, line 12:
```suggestion
        CanApply(c *Cluster) bool // replaces Validate?
	Apply(c *Cluster)
```

e2e/actions.go
```diff
@@ -0,0 +1,67 @@
...
+	remainingNodes := runningNodes - 1
+	if remainingNodes < maxFaultyNodes {
+		return fmt.Errorf("dropping a node would jeopardize Byzantine fault-tollerant conditions.\nExpected at least %d nodes to run, but action would leave it to %d",
+			maxFaultyNodes, remainingNodes)
+	}
+	return nil
+
+}
+
+func (dn *DropNodeAction) Apply(c *Cluster) {
```
ferranbt, line 35:
```suggestion
func (dn *DropNodeAction) Apply(c *Cluster) func() {
```

e2e/actions.go
```diff
@@ -0,0 +1,67 @@
...
+
+	maxFaultyNodes := pbft.MaxFaultyNodes(len(c.nodes))
+	runningNodes := c.GetRunningNodes()
+	log.Printf("BEGIN Running nodes: %v", runningNodes)
+
+	// Stop a node with some probability
+	nodeToStop := runningNodes[rand.Intn(len(runningNodes))]
+	log.Printf("Dropping node '%s'.", nodeToStop)
+	c.StopNode(nodeToStop.name)
+
```
ferranbt, line 49:
```
return func() {
     nodeToStop.Start() // this might need some locks
}
```

e2e/fuzz/runner.go
```diff
@@ -0,0 +1,115 @@
...
+func (r *Runner) Run(d time.Duration) error {
+	r.cluster.Start()
+	// TODO: we need to stop the cluster, what about tracer inside cluster stop?
+	// defer r.cluster.Stop()
+	done := time.After(d)
+	actionsApplied := make([]int, 0)
+	r.wg.Add(1)
+	go func() {
+		defer r.wg.Done()
+		for {
```
ferranbt, line 42:
```
var reverts []func()

for {
    select {
        case <-time.After(5 * time.Second):
             action := &NodeDropAction{} // use an array once more actions are available
             if action.CanApply(c) {
                  revertFn := action.Apply(c)
                  reverts = append(reverts, revertFn)
             }
        case <-time.After(3 * time.Second):
             if len(reverts) == 0 {
                  continue
             }
             if // 20% chance {
                 var revertFn func()
                 revertFn, reverts = reverts[0], reverts[1:] // pop a revert
                 revertFn()
             }                 
    }
}
```

ferranbt (changes requested):

e2e/fuzz/runner.go
Stefan-Ethernal: Should we randomize intervals for apply, revert (and possibly validating a cluster) as well or we shouldn't go that way? :slightly_smiling_face: 
However think that there might be some limitations (e.g. validation interval shouldn't be smaller than at least apply interval if not apply+revert interval etc.).

e2e/actions.go
```diff
@@ -0,0 +1,47 @@
...
+type RevertFunc func()
+
+type FunctionalAction interface {
+	CanApply(c *Cluster) bool
+	Apply(c *Cluster) RevertFunc
+}
+
+// Encapsulates logic for dropping nodes action.
+type DropNodeAction struct {
+	lock sync.Mutex
```
ferranbt, line 20: Why do we need this lock? Once the cluster has started (`Start()` called) this is the only place where `Start` and `Stop` are going to be called and they are already synced over the framework with the revert and apply channels.

ferranbt (changes requested):

e2e/actions.go
Stefan-Ethernal: In fact apparently we don't need it eventually. Although at some point you have suggested it (but now seems to me that you were referring to the logic contained within `Start` function: https://github.com/0xPolygon/pbft-consensus/pull/24#discussion_r810993635) and I thought that maybe we would spawn some additional routines for apply, revert and/or cluster validation, but seems that wouldn't be the case at the end, so we can remove lock.

ferranbt (approved):

PR #25, opened 2022-02-11
POS 204: Introduce scenarios support

Stefan-Ethernal:
#### Scenarios
This PR introduces scenarios support to the fuzz framework. Scenarios are array of predefined actions. Executing a scenario means applying each action in an array to the given node cluster.

#### Scenario generator
`ScenarioGenerator` is a component which generates an array of predefined number of scenarios, each containing given number of actions within it. Actions are randomly picked from the `actionRepository` (which holds all the available actions) and packed into the `Scenario` array. 

#### Runner
Fuzz `Runner` is altered in order to run scenarios instead of single `Action`.


Stefan-Ethernal:
Closing this PR, since we have gave up on this approach.

PR #26, opened 2022-02-15
Feat/validate proposal

ferranbt:
This PR does two things:
- It changes the Backend interface to remove the `Hash` function. Now, the Hash of the proposal is supplied during `BuildProposal` in case of a producer and it needs to get validated in `ValidateProposal` for a validator. This makes much more sense to me because before the hash felt like some transient data that was generated on the fly for specific purposes but now it is part of the state. Besides, we have much more flexibility on what that hash represents.
- Since the proposal hash is available on the state, we can ensure that only messages with the correct hash are being processed during Prepare and Commit stages.

### DIFF
--- a/state.go
+++ b/state.go
@@ -1,6 +1,7 @@
 package pbft
 
 import (
+	"bytes"
 	"fmt"
 	"sync/atomic"
 	"time"
@@ -43,8 +44,8 @@ type MessageReq struct {
 	// view is the view assigned to the message
 	View *View
 
-	// hash of the locked proposal
-	Digest string
+	// hash of the proposal (only for prepare messages)
+	Hash []byte
 
 	// proposal is the arbitrary data proposal (only for preprepare messages)
 	Proposal []byte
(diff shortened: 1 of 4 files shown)


state.go
```diff
@@ -129,6 +142,24 @@ type Proposal struct {
 
 	// Time is the time to create the proposal
 	Time time.Time
+
+	// Hash is the digest of the data to seal
+	Hash []byte
```
Stefan-Ethernal, line 147: Should we make `Hash` a read-only field, disallowing setting it from the outside, since it is derived type of data? 
I was thinking of having `NewProposal` function which would instantiate `Proposal` and set `Hash` field based on hashing `Data` field.

Also should we validate whether `Hash` is matching `hash(Data)` or we shouldn't mind about it (although equality is checked against Hash field, so that could be the same thing implicitly at the end)?

state.go
```diff
@@ -43,13 +44,25 @@ type MessageReq struct {
 	// view is the view assigned to the message
 	View *View
 
-	// hash of the locked proposal
-	Digest string
+	// hash of the proposal
+	Hash []byte
```
Stefan-Ethernal, line 48: Similair goes with `MessageReq` as with `Proposal` struct (see https://github.com/0xPolygon/pbft-consensus/pull/26/files#r808749656).

state.go
ferranbt: A couple of interesting points here. We cannot just do `hash=sha(Proposal)` because the hash has to exclude some fields of the proposal (this is because this hash is used to seal and validate the block but some fields of the block are generated after the block is sealed in pbft).

In order to solve that there was the `Hash` function in the backend that would take a proposal and hash it.

This PR does not change the fact that the backend has to be the one hashing the proposal, it only changes the place where it is done such that we can do a more meaningful validation. It is not the same to discard a block because the hash is wrong than let the pbft protocol fail silently because the Hash callback returned an incorrect Hash.

0xSasaPrsic (approved):

PR #27, opened 2022-02-15
Add validation for commit message

ferranbt:
This PR is a design idea (not to be merged yet).

We should aim to validate as early as possible on the Pbft protocol. For example, instead of validating if the transactions in the block are valid just at the `Commit` phase right at the end, we can do it on the `Validation` stage. The same applies for commit seal messages, we can check them right when we receive them.

Ideally, the goal should be that `backend.Insert` does not have any situation to fail since everything that can spawn an error has been found beforehand.

### DIFF
--- a/consensus.go
+++ b/consensus.go
@@ -118,6 +118,9 @@ type Backend interface {
 
 	// IsStuck returns whether the pbft is stucked
 	IsStuck(num uint64) (uint64, bool)
+
+	// ValidateCommit is used to validate that a given commit is valid
+	ValidateCommit(seal []byte) error
 }
 
 // Pbft represents the PBFT consensus mechanism object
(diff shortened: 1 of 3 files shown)


Stefan-Ethernal (approved):
I'm good with the idea, although I'm currently unsure how sealed/committed proposal validation should look like, but that's different topic I suppose.

0xSasaPrsic (approved):

ferranbt:
[This](https://github.com/0xPolygon/polygon-edge/blob/bd192e4677a2b15e67765c3949c13d1e43d945eb/consensus/ibft/sign.go#L168) is how the verification is done in polygon-sdk. The seal is the ecdsa signature of the hash (the one that locks the proposal). Then, it is just a matter of ecrecover and make sure that the address matches the sender.

However, now that I think about it, it also needs to include the sender to verify since the backend will already know the hash.

0xSasaPrsic (approved):

PR #30, opened 2022-02-22
Fuzz partitioning action.

0xSasaPrsic:
Implementation of fuzz partition action that will create majority and  minority partitions. Majority partition is a subset of nodes that can create a block.

### DIFF
--- a/e2e/fuzz/runner.go
+++ b/e2e/fuzz/runner.go
@@ -91,8 +91,12 @@ func validateNodes(c *e2e.Cluster) {
 		currentHeight := c.GetMaxHeight(runningNodes)
 		expectedHeight := currentHeight + 10
 		log.Printf("Current height %v and waiting expected %v height.\n", currentHeight, expectedHeight)
-		err := c.WaitForHeight(expectedHeight, 1*time.Minute, runningNodes)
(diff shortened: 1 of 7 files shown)


e2e/fuzz/runner.go
```diff
@@ -104,29 +108,52 @@ func validateNodes(c *e2e.Cluster) {
...
-	totalNodesCount := 0
+	totalNodesCount := len(c.Nodes())
 	var runningNodes []string
-	for _, n := range c.Nodes() {
-		if n.IsRunning() {
+	var partitions map[string][]string
+	// running nodes in majority partition
+	hook := c.GetTransportHook()
+	if hook != nil {
+		partitions = hook.GetPartitions()
```
ferranbt, line 119: If we have a new attack like the flow map used for the liveness issue (though not sure if we will use that), we cannot have anymore this type of check. I think it does work now but we have to figure out how to scale this for any arbitrary type of partition.

ferranbt (approved):

e2e/fuzz/runner.go
0xSasaPrsic: Yes, we would need to design the flow map and then to find better way of checking if there is enough nodes connected. Current implementation relays on the fact that all nodes inside partition are connected.

PR #34, opened 2022-03-09
Add info to Init

ferranbt:
In #18, an `Init` function was introduced to signal when a new round was starting. After integrating it in v3 I realized that it would be more helpful if this function also returns some metadata/info about the round.

### DIFF
--- a/consensus.go
+++ b/consensus.go
@@ -114,10 +114,17 @@ type Backend interface {
 	Hash(p []byte) []byte
 
 	// Init is used to signal the backend that a new round is going to start.
-	Init()
+	Init(*RoundInfo)
 
 	// IsStuck returns whether the pbft is stucked
 	IsStuck(num uint64) (uint64, bool)
+}
+
+// RoundInfo is the information about the round
+type RoundInfo struct {
+	IsProposer bool
+	Proposer   NodeID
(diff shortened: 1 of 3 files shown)


consensus.go
```diff
@@ -114,12 +114,19 @@ type Backend interface {
...
+	Init(*RoundInfo)
 
 	// IsStuck returns whether the pbft is stucked
 	IsStuck(num uint64) (uint64, bool)
 }
 
+// RoundInfo is the information about the round
+type RoundInfo struct {
+	IsProposer bool
+	Proposer   NodeID
```
Stefan-Ethernal, line 126: Do we need both `IsProposer` and `Proposer` fields? Looks like they are a bit redundant, although not sure about usage context, maybe it is easier to have both information in place...

Stefan-Ethernal (approved):

0xSasaPrsic (approved):

consensus.go
ferranbt: I think it does not hurt to have both, because the `Backend` implementation does not necessary need to have the local NodeID to check.

Stefan-Ethernal: Ok agreed!

PR #35, opened 2022-03-11
Simplify e2e network sync

0xSasaPrsic:
Implementation of simplified node sync. Node is persisting only index up to which is in sync with the network. All finalized blocks and last proposer are persisted in the cluster (`sealedProposals`, `proposer`). Synchronization is done by checking current height of the network and setting node `syncIndex`.

### DIFF
--- a/e2e/framework.go
+++ b/e2e/framework.go
@@ -70,7 +70,7 @@ type cluster struct {
 }
 
 // getSyncIndex returns a index up to which the node is synced with the network
-func (c *cluster) getSyncIndex(node string) uint64 {
+func (c *cluster) getSyncIndex(node string) int64 {
 	return c.nodes[node].getSyncIndex()
 }
 
@@ -80,13 +80,14 @@ func (c *cluster) insertFinalProposal(p *pbft.SealedProposal) {
 	defer c.lock.Unlock()
 
 	lastIndex := len(c.sealedProposals) - 1
-	insertIndex := p.Number - 2
+	insertIndex := p.Number - 1
 	if insertIndex == uint64(lastIndex) {
 		// already exists
 		if !c.sealedProposals[insertIndex].Proposal.Equal(p.Proposal) {
 			panic("Proposals are not equal")
 		}
 	} else {
+		fmt.Printf("Proposer: %v Insert height %v\n", p.Proposer, p.Number)
 		c.sealedProposals = append(c.sealedProposals, p)
 	}
 }
@@ -141,9 +142,7 @@ func (c *cluster) IsStuck(timeout time.Duration, nodes ...[]string) {
 	nodeHeight := map[string]uint64{}
 	isStuck := func() bool {
 		for _, n := range queryNodes {
-			c.lock.Lock()
-			height := c.getNodeHeight(c.nodes[n])
-			c.lock.Unlock()
+			height := c.nodes[n].getNodeHeight()
 			if lastHeight, ok := nodeHeight[n]; ok {
 				if lastHeight != height {
 					return false
@@ -180,7 +179,7 @@ func (c *cluster) WaitForHeight(num uint64, timeout time.Duration, nodes ...[]st
 		defer c.lock.Unlock()
 
 		for _, name := range queryNodes {
-			if c.getNodeHeight(c.nodes[name]) < num {
+			if c.nodes[name].getNodeHeight() < num {
 				return false
 			}
 		}
@@ -200,21 +199,16 @@ func (c *cluster) WaitForHeight(num uint64, timeout time.Duration, nodes ...[]st
 	}
 }
 
-func (c *cluster) getNodeHeight(n *node) uint64 {
-	index := n.getSyncIndex()
-	height := uint64(1) // initial height is always 1 since 0 is the genesis
-	if len(c.sealedProposals) != 0 {
-		height = c.sealedProposals[index].Number
-	}
-	return height
+func (n *node) getNodeHeight() uint64 {
+	return uint64(n.getSyncIndex()) + 1
 }
 
-func (c *cluster) syncWithNetwork(nodeID string) (uint64, uint64) {
+func (c *cluster) syncWithNetwork(nodeID string) (uint64, int64) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
 	var height uint64
-	var syncIndex uint64
+	var syncIndex int64 = -1
 	for _, n := range c.nodes {
 		if n.name == nodeID {
 			continue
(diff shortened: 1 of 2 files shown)


e2e/framework.go
```diff
@@ -60,10 +60,73 @@ func initTracer(name string) *sdktrace.TracerProvider {
...
+	}
+	return height
+}
+
+// insertFinalProposal inserts final proposal from the node to the cluster and returns index up to which the node is synced
+func (c *cluster) insertFinalProposal(p *pbft.SealedProposal) uint64 {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	for _, v := range c.sealedProposals {
```
ferranbt, line 107: This is the safe check to validate that two proposals are the same. If all the nodes block at this function, shouldn't the proposal be always the last one?

e2e/framework.go
```diff
@@ -60,10 +60,73 @@ func initTracer(name string) *sdktrace.TracerProvider {
...
-	nodes  map[string]*node
-	tracer *sdktrace.TracerProvider
-	hook   transportHook
+	t               *testing.T
+	lock            sync.Mutex
+	nodes           map[string]*node
+	tracer          *sdktrace.TracerProvider
+	hook            transportHook
+	sealedProposals []*pbft.SealedProposal
+	proposer        pbft.NodeID
```
ferranbt, line 69: Rename to `lastProposer`.

e2e/framework.go
```diff
@@ -60,10 +60,73 @@ func initTracer(name string) *sdktrace.TracerProvider {
...
+}
+
+// getCurrentIndex returns a last index up to which there are inserted proposals
+func (c *cluster) getCurrentIndex() uint64 {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	idx := 0
+	if len(c.sealedProposals) > 0 {
```
ferranbt, line 77: So if there are no items the index is '0' and if there is only one item the index is also 0?

e2e/framework.go
```diff
@@ -193,6 +237,46 @@ func (c *cluster) WaitForHeight(num uint64, timeout time.Duration, nodes ...[]st
 	}
 }
 
+func (c *cluster) syncWithNetwork(ourselves string) (uint64, uint64) {
```
ferranbt, line 240: Can we rename `ourselves` to `nodeID`? It gives the impression that the string represents something in the cluster.

ferranbt (changes requested):

ferranbt:
Everything is looking good on the draft.

e2e/framework.go
0xSasaPrsic: Removed since it is no longer required.

e2e/framework.go
```diff
@@ -193,6 +204,58 @@ func (c *cluster) WaitForHeight(num uint64, timeout time.Duration, nodes ...[]st
 	}
 }
 
+func (c *cluster) getNodeHeight(n *node) uint64 {
```
ferranbt, line 207: Now that I think about this, do we really need to get the proposal to get the Height? Shouldn't the height be always equal to the index? I think we can remove a noisy lock if we do it that way.

e2e/framework.go
```diff
@@ -341,19 +385,17 @@ func (n *node) Start() {
...
-		n.proposals = history
-
+		_, syncIndex := n.c.syncWithNetwork(n.name)
+		n.setSyncIndex(syncIndex)
 		for {
 			fsm := &fsm{
 				n:            n,
 				nodes:        n.nodes,
-				lastProposer: n.lastProposer(),
+				lastProposer: n.c.getLastProposer(),
```
ferranbt, line 394: I think this is not totally correct, what if due to partitions you do not sync up to the tip of the chain? You should get the proposal at your index and get the proposer from there. I would remove `getLastProposer` and do a `getProposalByIndex(indx uint64) *pbft.Proposal`. With this we can even remove `lastProposer` variable.

e2e/framework.go
```diff
@@ -60,10 +61,37 @@ func initTracer(name string) *sdktrace.TracerProvider {
...
+func (c *cluster) getSyncIndex(node string) uint64 {
+	return c.nodes[node].getSyncIndex()
+}
+
+// insertFinalProposal inserts final proposal from the node to the cluster and returns index up to which the node is synced
+func (c *cluster) insertFinalProposal(p *pbft.SealedProposal) uint64 {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if len(c.sealedProposals) > 0 {
+		v := c.sealedProposals[len(c.sealedProposals)-1]
```
ferranbt, line 83: Sorry, I know that I already made a comment here but I though about something different. Could there be a situation of many nodes in a fast pbft protocol in which some subset with quorum start another round and finish before another set finishes writing?

I think it is safer to assume that `p.Number` == your index in `c.sealedProposals` and do that check instead.

```
lastIndex := len(c.sealedProposals)
if lastIndex + 1 == p.Number {
   // new proposal
   c.sealedProposals = append(c.sealedProposals, p)
} else {
   // already in, validate
  if !p.Equal(c.sealedProposals[p.Number] {
     panic(...)
  }
}
```

e2e/framework.go
```diff
@@ -285,49 +351,27 @@ func (n *node) isStuck(num uint64) (uint64, bool) {
...
-	res = append(res, n.proposals...)
-
-	number := uint64(0)
-	if len(res) != 0 {
-		number = uint64(res[len(res)-1].Number)
-	}
-	return number, res
+func (n *node) Insert(pp *pbft.SealedProposal) error {
+	index := n.c.insertFinalProposal(pp)
+	n.setSyncIndex(index)
```
ferranbt, line 356: I am not sure we need to set the index here, it would be better to do it at the end of `pbft.Run(ctx)`. At least in the `v3` repo, we change the current block after the protocol is over. Thus, we can also remove `insertFinalProposal` returning an index.

ferranbt (changes requested):

e2e/framework.go
0xSasaPrsic: Height is not equal to index because, initial height is always 1 since 0 us genesis, 
First inserted height would be 2 on index 0. We could do some kind of calculation because difference between height and index is 2.

e2e/framework.go
0xSasaPrsic: Yes, I agree, this way we don't need additional `lastProposer` field in the cluster and also it will use `localIndex` to start sync from there.

e2e/framework.go
0xSasaPrsic: Do you think we should update local index once the state machine goes to `done` state? Here sync index is updated once the proposal is finalised.

ferranbt: Yes, update the index after `done` state. Then, if is enough to do `n.setSyncIndex(currentIndex+1)`.

ferranbt (approved):

PR #37, opened 2022-03-21
Fixed partitions reset issue when node gets isolated

goran-ethernal:
Fixed the issue on partitions reset when reverting `PartitionAction`, where a node gets isolated from the rest of the network and stops communicating with other nodes.

When calling Reset on a partition, `PartitionTransport` subsets were set to an empty map, which can lead to isolation of a given node, since `Gossip` will check if a node is connected to another node, and since its subset is empty it will return false and node will get stuck.

This is solved by setting subsets map on `PartitionTransport` to `nil` on `Reset`, and `nil` is handled in `isConnected` method as a valid case which means that given node is connected to the entire network.
When another `PartitionAction` is created it will set it its map to another partition and action can proceed to execute.

**NOTE** For flow map, given behavior will be implemented and pushed when it gets merged with this branch.

### DIFF
--- a/e2e/transport.go
+++ b/e2e/transport.go
@@ -88,6 +88,10 @@ func newPartitionTransport(jitterMax time.Duration) *partitionTransport {
 }
 
 func (p *partitionTransport) isConnected(from, to pbft.NodeID) bool {
+	if p.subsets == nil {
+		return true
+	}
+
 	subset, ok := p.subsets[string(from)]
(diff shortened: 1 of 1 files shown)


Stefan-Ethernal (approved):
LGTM, but let's wait for Ferran's approval.

ferranbt (approved):

PR #41, opened 2022-03-25
Prevent insertion of duplicate sealed proposals on different indices

Stefan-Ethernal:
### Introduction and context
Different nodes (the ones which were found on some of the previous sequences) were inserting same sealed proposals on different indices within the sealed proposals slice. 
Major part of making replay messages work properly was down to this problem. It was the reason why proposer calculation was incorrect when performing replay messages (last proposer is calculated based on last inserted sealed proposal, which holds information about proposer as well whereas current proposer is `indexOf(lastProposer) + round + 1`).

### Explanation
Since all the nodes are sharing sealed proposals, which is cached on a single place, within the cluster (changes made through PR https://github.com/0xPolygon/pbft-consensus/pull/35) and due to the fact that different nodes can found themselves on different sequences, nodes were trying to insert repeated sealed proposals on different index within a `sealedProposals` slice, held by `e2e.Cluster`. This issue occurs when some of the nodes have enough messages to insert a sealed proposal for `sequence+1`, whereas others are finalizing a sealed proposal for `sequence`.

This was the case because of logical condition for checking whether same proposal already exists as shown in snippet below (`insertIndex == uint64(lastIndex)`), which was checking whether proposal already exists only if insertion index is the same as last index:

```Go
lastIndex := len(c.sealedProposals) - 1
insertIndex := p.Number - 1
if insertIndex == uint64(lastIndex) {
	// already exists
	if !c.sealedProposals[insertIndex].Proposal.Equal(p.Proposal) {
		panic("Proposals are not equal")
	}
	pbft.Log(fmt.Sprintf("Proposal repeated %+v. InsertIndex=%d, LastIndex=%d\n", p, insertIndex, lastIndex))
}
```

#### Example
We have a cluster of 3 nodes: {N0, N1, N2} and all the messages for sequences 1 and 2 are already present within the message queues for all the nodes (which is the case in [replay messages](https://github.com/0xPolygon/pbft-consensus/pull/38) feature). 

N0 is the proposer for the sequence 1 and it gossips PRE-PREPARE and PREPARE messages to N1 and N2. N1 and N2 respond to N0 and each other with PREPARE (and COMMIT) messages, so they are eligible to seal a proposal and insert it, which they do. Only one of them will do so, since the second one, a condition which is checking if proposal already exists (`insertIndex == uint64(lastIndex)`) is going to be triggered. At this moment sealed proposals looks like this (let's imagine for simplicity that proposal data will always contain sequence which corresponds to the one when the proposal is built):
Sealed Proposals=[Seq=1 Proposer=N0 Proposal=[1]].

N1 and N2 advances to sequence 2. N1 is proposer and it relay PRE-PREPARE and PREPARE messages with N2, so once they have relayed enough PREPARE (and COMMIT) messages, each of these two will try to insert sealed proposal for sequence 2, which they do, but again, only 1st one who attempts will be successful. Sealed proposals slice looks like this:
Sealed Proposals=[Seq=1 Proposer=N0 Proposal=[1], Seq=2 Proposer=N1, Proposal=[2]].
At this point N0 manages to insert sealed proposal for sequence 1 and it will eventually manage to append it to the end, since `lastIndex = 2 - 1 = 1` and `insertIndex = 1 - 1 = 0` and `insertIndex != lastIndex`, which means that else branch is getting executed. Sealed proposals at this point is like this (bold one sealed proposal is repeated):
Sealed Proposals=[Seq=1 Proposer=N0 Proposal=[1], Seq=2 Proposer=N1 Proposal=[2], **Seq=1 Proposer=N0 Proposal=[1]**]. This means that last proposer will be wrong for some of next sequence (either 3 or 4, although most often 4, because at this point N1 and N2 are already processing sequence 3). 

### Resolution proposal
Since nodes are running independently, we cannot rely that if insertion index is different than last inserted index, proposals aren't repeated. Therefore we should check if proposals are the same even if insertion index is less than or equal to last inserted index. Also, whenever a sealed proposal is being inserted, we should iterate through whole sealed proposals slice and detect whether there is the one which holds same data and is already inserted.

### DIFF
--- a/e2e/framework.go
+++ b/e2e/framework.go
@@ -102,19 +102,24 @@ func (c *cluster) getSyncIndex(node string) int64 {
 }
 
 // insertFinalProposal inserts final proposal from the node to the cluster
-func (c *cluster) insertFinalProposal(p *pbft.SealedProposal) {
+func (c *cluster) insertFinalProposal(sealProp *pbft.SealedProposal) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
 	lastIndex := len(c.sealedProposals) - 1
-	insertIndex := p.Number - 1
-	if insertIndex == uint64(lastIndex) {
+	insertIndex := sealProp.Number - 1
+	if len(c.sealedProposals) > 0 && insertIndex <= uint64(lastIndex) {
 		// already exists
-		if !c.sealedProposals[insertIndex].Proposal.Equal(p.Proposal) {
-			panic("Proposals are not equal")
+		if !c.sealedProposals[insertIndex].Proposal.Equal(sealProp.Proposal) {
+			panic("Existing proposal on a given position is not not equal to the one being inserted to the same position")
 		}
 	} else {
-		c.sealedProposals = append(c.sealedProposals, p)
+		for _, currentSealProp := range c.sealedProposals {
+			if currentSealProp.Proposal.Equal(sealProp.Proposal) {
+				panic("Proposal already exists")
+			}
+		}
+		c.sealedProposals = append(c.sealedProposals, sealProp)
 	}
 }
 
--- a/e2e/framework.go
+++ b/e2e/framework.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"crypto/sha1"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


e2e/framework.go
```diff
@@ -102,19 +102,24 @@ func (c *cluster) getSyncIndex(node string) int64 {
...
+	if len(c.sealedProposals) > 0 && insertIndex <= uint64(lastIndex) {
 		// already exists
-		if !c.sealedProposals[insertIndex].Proposal.Equal(p.Proposal) {
-			panic("Proposals are not equal")
+		if !c.sealedProposals[insertIndex].Proposal.Equal(sealProp.Proposal) {
+			panic("Existing proposal on a given position is not not equal to the one being inserted to the same position")
 		}
 	} else {
-		c.sealedProposals = append(c.sealedProposals, p)
+		for _, currentSealProp := range c.sealedProposals {
```
ferranbt, line 117: So this is only a safe check to make sure that we do not add a proposal twice?

e2e/framework.go
```diff
@@ -102,19 +102,24 @@ func (c *cluster) getSyncIndex(node string) int64 {
...
+			if currentSealProp.Proposal.Equal(sealProp.Proposal) {
+				panic("Proposal already exists")
+			}
+		}
+		c.sealedProposals = append(c.sealedProposals, sealProp)
```
ferranbt, line 122: Is there any check to make sure that the proposals are being included in linear order?

ferranbt (changes requested):

e2e/framework.go
Stefan-Ethernal: Yes, although not sure whether we should keep it or not. However, it is not expected to have same proposal inserted more than once in real world.
Btw ,existing fuzz tests (not the fuzz framework, but the fuzz tests currently running on GH), failed at this check when I've originally created PR (https://github.com/0xPolygon/pbft-consensus/runs/5688158137?check_suite_focus=true). Although since we don't have verbose logging enabled for fuzz tests running, I cannot see which test failed and why. Re-running of tests was successful and I wasn't able to replicate it on my machine as well.

e2e/framework.go
Stefan-Ethernal: Currently not, but that's good point, we should add checking whether currently inserted sealed proposal's sequence is last added sequence incremented by one.

e2e/framework.go
ferranbt: Do you mean the e2e tests? Then, let's keep the check to see if it fails again.

I am afraid that if we run the fuzz test for a long time looping over this array all the time becomes a bottleneck.

Stefan-Ethernal: > Do you mean the e2e tests? 

I mean the subset of e2e tests which are marked as fuzz (e.g. TestFuzz_NetworkChurn and TestFuzz_Unreliable_Network).

> I am afraid that if we run the fuzz test for a long time looping over this array all the time becomes a bottleneck.

Talking with Saša about it and maybe we could introduce an optimization, iterating just through subset of `sealedProposal` slice, although that wouldn't be bulletproof, and it would practically mean checking only last proposal most of the time, so it is not so meaningful.
`for _, currentSealProp := range c.sealedProposals[sealProp.Number - lastIndex:]`

Maybe we should eventually remove that check, since this is integration framework and proposal building is part of `Backend` mock, so if it isn't correct, it would mean that 99% is some problem with the integration framework. WDYT?

ferranbt (approved):

Stefan-Ethernal:
@ferranbt There is a small catch here: currently unit test, which is implemented as part of this PR (`Test_ClusterInsertFinalProposal`) and located in `framework_test.go` (within e2e folder) is being executed as part of e2e tests, instead of unit tests. 
I was considering separating and grouping tests using build tags and submitting those changes via a separate PR. What is your opinion? Is there some better way to address this?

PR #42, opened 2022-03-25
Add verbose logging

0xSasaPrsic:
Add verbose logging of the state machine and messages when reading from a queue.

### DIFF
--- a/consensus.go
+++ b/consensus.go
@@ -735,16 +735,16 @@ func (p *Pbft) getNextMessage(span trace.Span, timeout time.Duration) (*MessageR
 	for {
 		msg, discards := p.msgQueue.readMessageWithDiscards(p.getState(), p.state.view)
 		// send the discard messages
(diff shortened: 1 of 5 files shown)


consensus.go
```diff
@@ -735,13 +735,16 @@ func (p *Pbft) getNextMessage(span trace.Span, timeout time.Duration) (*MessageR
 	for {
 		msg, discards := p.msgQueue.readMessageWithDiscards(p.getState(), p.state.view)
 		// send the discard messages
+		p.logger.Printf("[DEBUG] Current state %s, number of prepared messages: %d, number of committed messages %d", PbftState(p.state.state), p.state.numPrepared(), p.state.numCommitted())
```
ferranbt, line 738:
```suggestion
		p.logger.Printf("[TRACE] Current state %s, number of prepared messages: %d, number of committed messages %d", PbftState(p.state.state), p.state.numPrepared(), p.state.numCommitted())
```

ferranbt (changes requested):
Looks good to me, could you change the level of the logs to TRACE instead?

ferranbt (approved):

PR #44, opened 2022-03-30
Improve stopping of nodes after finishing all messages

goran-ethernal:
Improved stopping of nodes on replay after they reach their last sequence in file.

Now replay will wait for nodes to read and process all the messages that were in the .flow file for their last sequence, and then will stop nodes as they reach the end of their execution.

This way we do not need to wait for continuous `Round Change`, and we are certain that all messages from the last sequence were processed.

### DIFF
--- a/e2e/fuzz/replay/replay_message_reader.go
+++ b/e2e/fuzz/replay/replay_message_reader.go
@@ -2,6 +2,7 @@ package replay
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -14,19 +15,23 @@ import (
 )
 
 const (
-	maxCharactersPerLine   = 2048 * 1024 // Increase Scanner buffer size to 2MB per line
-	roundMessagesWaitLimit = 50
-	messageChunkSize       = 200
+	maxCharactersPerLine = 2048 * 1024 // Increase Scanner buffer size to 2MB per line
+	messageChunkSize     = 200
 )
+
+type sequenceMessages struct {
+	sequence uint64
+	messages []*pbft.MessageReq
+}
 
 // replayMessageReader encapsulates logic for reading messages from flow file
 type replayMessageReader struct {
 	lock                   sync.Mutex
 	file                   *os.File
 	scanner                *bufio.Scanner
-	lastSequence           uint64
 	msgProcessingDone      chan string
 	nodesDoneWithExecution map[pbft.NodeID]bool
+	lastSequenceMessages   map[pbft.NodeID]*sequenceMessages
 }
 
 // openFile opens the file on provided location
@@ -74,12 +79,19 @@ func (r *replayMessageReader) readNodeMetaData() ([]string, error) {
 // readMessages reads messages from open .flow file and pushes them to appropriate nodes
 func (r *replayMessageReader) readMessages(cluster *e2e.Cluster) {
 	nodes := cluster.GetNodesMap()
-	r.nodesDoneWithExecution = make(map[pbft.NodeID]bool, len(nodes))
+
+	nodesCount := len(nodes)
+	r.nodesDoneWithExecution = make(map[pbft.NodeID]bool, nodesCount)
+	r.lastSequenceMessages = make(map[pbft.NodeID]*sequenceMessages, nodesCount)
 
 	messagesChannel := make(chan []*ReplayMessage)
 	doneChannel := make(chan struct{})
 
 	r.startChunkReading(messagesChannel, doneChannel)
+	nodeMessages := make(map[pbft.NodeID]map[uint64][]*pbft.MessageReq, nodesCount)
+	for _, n := range nodes {
+		nodeMessages[pbft.NodeID(n.GetName())] = make(map[uint64][]*pbft.MessageReq)
+	}
 
 	isDone := false
 LOOP:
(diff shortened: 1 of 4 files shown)


e2e/fuzz/replay/replay_message_reader.go
```diff
@@ -135,28 +160,59 @@ func (r *replayMessageReader) startChunkReading(messagesChannel chan []*ReplayMe
...
+		if messageIndexToRemove != -1 {
+			lastSequenceMessages.messages = append(lastSequenceMessages.messages[:messageIndexToRemove], lastSequenceMessages.messages[messageIndexToRemove+1:]...)
+		}
+
+		return len(lastSequenceMessages.messages) == 0
+	}
+}
+
+// areMessagesEqual compares if two messages are equal
+func areMessagesEqual(msgOne, msgTwo *pbft.MessageReq) bool {
```
Stefan-Ethernal, line 206: Think that it would be nicer if this was the method in MessageReq struct.

e2e/fuzz/replay/replay_message_reader.go
```diff
@@ -135,28 +160,59 @@ func (r *replayMessageReader) startChunkReading(messagesChannel chan []*ReplayMe
...
+	lastSequenceMessages := r.lastSequenceMessages[validatorId]
+
+	if len(lastSequenceMessages.messages) == 0 {
+		return true
+	} else {
+		messageIndexToRemove := -1
+		for i, message := range lastSequenceMessages.messages {
+			if areMessagesEqual(msg, message) {
+				messageIndexToRemove = i
+			}
```
Stefan-Ethernal, line 194: Add a break from the loop?

e2e/fuzz/replay/replay_message_reader.go
```diff
@@ -135,28 +160,59 @@ func (r *replayMessageReader) startChunkReading(messagesChannel chan []*ReplayMe
...
+		}
+
+		if messageIndexToRemove != -1 {
+			lastSequenceMessages.messages = append(lastSequenceMessages.messages[:messageIndexToRemove], lastSequenceMessages.messages[messageIndexToRemove+1:]...)
+		}
+
+		return len(lastSequenceMessages.messages) == 0
+	}
```
Stefan-Ethernal, line 202: Small suggestion in order to spare a line of code or so :slightly_smiling_face: 

```suggestion
        lastSeqMsgsCount := len(lastSequenceMessages.messages)
	if lastSeqMsgsCount > 0 {
		messageIndexToRemove := -1
		for i, message := range lastSequenceMessages.messages {
			if areMessagesEqual(msg, message) {
				messageIndexToRemove = i
			}
		}

		if messageIndexToRemove != -1 {
			lastSequenceMessages.messages = append(lastSequenceMessages.messages[:messageIndexToRemove], lastSequenceMessages.messages[messageIndexToRemove+1:]...)
		}
	}
        return lastSeqMsgsCount == 0
```

e2e/fuzz/replay/replay_message_reader.go
```diff
@@ -135,28 +160,59 @@ func (r *replayMessageReader) startChunkReading(messagesChannel chan []*ReplayMe
...
+// processMessage checks if message is a timeout, or if all messages have been read from the queue
+func (r *replayMessageReader) processMessage(validatorId pbft.NodeID, msg *pbft.MessageReq) bool {
+	if isTimeoutMessage(msg) {
+		return true
+	}
+
+	if msg.View.Sequence > r.lastSequenceMessages[validatorId].sequence ||
+		(msg.View.Sequence == r.lastSequenceMessages[validatorId].sequence && r.areMessagesFromLastSequenceProcessed(msg, validatorId)) {
 		r.lock.Lock()
-		if _, isDone := r.nodesDoneWithExecution[validatorId]; !isDone {
```
Stefan-Ethernal, line 145: Why did you give up on this pattern? Code looks more compact that way.

e2e/fuzz/replay/replay_message_reader.go
```diff
@@ -135,28 +160,59 @@ func (r *replayMessageReader) startChunkReading(messagesChannel chan []*ReplayMe
...
 }
 
-// checks if given node is done with execution of messages
-func (r *replayMessageReader) checkIfDoneWithExecution(validatorId pbft.NodeID, msg *pbft.MessageReq) {
-	//when nodes reach sequence that is higher than the sequence in file,
-	//or if it is in constant round change we know we either reached the end of execution or nodes are stuck and there is a problem
-	if msg.View.Sequence > r.lastSequence ||
-		isSequenceInContinuousRoundChange(msg, validatorId) {
+// processMessage checks if message is a timeout, or if all messages have been read from the queue
+func (r *replayMessageReader) processMessage(validatorId pbft.NodeID, msg *pbft.MessageReq) bool {
```
Stefan-Ethernal, line 164: I would still prefer to extract this part of checking whether is it a timeout message or not to the replay_message_notifier (where it was prior to this change) and rename this one to markMessageProcessingDone (or something like that) and it doesn't need to return anything as a result. It would just mark node as done within the map if needed and that's all.
My rationale is that this method should do exactly one thing: mark message processing as done if necessary and that's it.

e2e/fuzz/replay/replay_message_reader.go
```diff
@@ -135,28 +160,59 @@ func (r *replayMessageReader) startChunkReading(messagesChannel chan []*ReplayMe
...
+	if isTimeoutMessage(msg) {
+		return true
+	}
+
+	if msg.View.Sequence > r.lastSequenceMessages[validatorId].sequence ||
+		(msg.View.Sequence == r.lastSequenceMessages[validatorId].sequence && r.areMessagesFromLastSequenceProcessed(msg, validatorId)) {
 		r.lock.Lock()
-		if _, isDone := r.nodesDoneWithExecution[validatorId]; !isDone {
+		_, isDone := r.nodesDoneWithExecution[validatorId]
+		if !isDone {
```
Stefan-Ethernal, line 173: Why did you give up on this pattern? Code looks more compact that way (at least to me) :slightly_smiling_face: 
```suggestion
		if _, isDone := r.nodesDoneWithExecution[validatorId]; !isDone {
```

e2e/fuzz/replay/replay_message_reader.go
```diff
@@ -92,12 +104,25 @@ LOOP:
...
-					if r.lastSequence < message.Message.View.Sequence {
-						r.lastSequence = message.Message.View.Sequence
-					}
+					nodeMessages[message.To][message.Message.View.Sequence] = append(nodeMessages[message.To][message.Message.View.Sequence], message.Message)
 				}
 			}
 		case <-doneChannel:
+			for name, n := range nodeMessages {
+				var nodeLastSequence uint64
+				nodeLastSequence = 0
```
Stefan-Ethernal, line 113:
```suggestion
				nodeLastSequence := uint64(0)
```

e2e/fuzz/replay/replay_message_reader.go
goran-ethernal: since we change the count of messages in if messageIndexToRemove != -1, we can not use the lastSeqMsgsCount variable like this. But will use something from suggestion.

ferranbt (dismissed):

Stefan-Ethernal (approved):

Stefan-Ethernal (approved):

PR #45, opened 2022-03-31
Build propsal refactor

goran-ethernal:
Mock implementation of `Backend` interface in integration testing framework is currently always returning current block height for proposal, once `BuildProposal` function is invoked. 

Now `BuildProposal` is changed to return random `byte array` and replay messages implements its own backend with `BuildProposal` mechanism, where it assures that same proposal is built as the one being injected via `PRE-PREPARE` message for given sequence in `.flow` file (otherwise it turns out that proposal which proposer sets to its internal state differs than the one it is gossiping around).

### DIFF
--- a/e2e/framework.go
+++ b/e2e/framework.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"math/rand"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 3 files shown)


Stefan-Ethernal (approved):

e2e/framework.go
```diff
@@ -618,7 +634,16 @@ func (f *fsm) ValidatorSet() pbft.ValidatorSet {
 	return &vv
 }
 
-func hash(p []byte) []byte {
+// SetBackendData implements IntegrationBackend interface and sets the data needed for backend
+func (fsm *Fsm) SetBackendData(n *node) {
```
0xSasaPrsic, line 638: name receiver should be the same as in the rest of the methods.

ferranbt (approved):

PR #48, opened 2022-04-05
Drop node simulation in replay

goran-ethernal:
Replay messages feature now saves actions that occurred in fuzz run to a separate .flow file so that they can be simulated on replay as well. For now, only Drop Node actions are saved to .flow file and simulated in replay.

When running `fuzz-run` command, messages are saved to `messages.flow` file, but other meta data (node names, actions, last sequences) are saved to `metaData.flow` file inside the **SavedState** folder.

`metaData.flow` file contains data about:

- node names - name of nodes that were created on cluster.
- last sequences in which node were when cluster was stopped - this is needed for replay to know exactly when to stop node from executing in replay (this way we now, that the node reached its end and it needs to be stopped).
- actions - data about drop node actions that occurred in fuzz run, as well as their revert actions.

Look of `messages.flow` file remains the same. `metaData.flow` file, looks as follows:
```
["NODE_0","NODE_1","NODE_2","NODE_3","NODE_4"]
{"actionType":"LastSequence","data":"NODE_0","sequence":10,"round":0}
{"actionType":"LastSequence","data":"NODE_1","sequence":11,"round":0}
{"actionType":"LastSequence","data":"NODE_2","sequence":11,"round":0}
{"actionType":"LastSequence","data":"NODE_3","sequence":11,"round":0}
{"actionType":"LastSequence","data":"NODE_4","sequence":11,"round":0}
{"actionType":"DropNode","data":"NODE_3","sequence":5,"round":0}
{"actionType":"RevertDropNode","data":"NODE_3","sequence":6,"round":0}
```
`MetaData` struct is used to store all necessary data to `metaData.flow` file. On loading the given file, based on `actionType` property, replay knows in which map to store a given action.

`replay-messages` command is now changed to receive folder path where `messages.flow` and to `metaData.flow` files are stored, e.g:
`go run ./e2e/fuzz/cmd/main.go replay-messages -filesDirectory=../SavedData`

**NOTE**: both files are needed for replay to execute.

Simulating node actions and stopping nodes when they are done with execution is now handled in `replay_node_execution.go`.
Once replay is started, when node reads its next message from queue, it can now recognize if it needs to shut down properly or if it needs to simulate a drop. 
We know that a node needs to be stopped when it has nothing to read from the queue and it reached the sequence and round that is defined in its `LastSequence` in `metaData.flow` file.
We know that a node needs to be dropped if it has nothing to read from the queue for given sequence and round, but given sequence and round are not the same as the one defined in its `LastSequence` in `metaData.flow` file.

`replayNodeExecutionHandler` starts a go routine before starting the cluster that listens if a node needs to be dropped, or if its done with execution, or if it needs to be restarted when a sequence is reached in cluster that corresponds to sequence in which node drop was reverted as defined in `RevertNode` action in `metaData.flow` file.

Once all nodes are done with execution, `replayNodeExecutionHandler` will stop the cluster, and command will exit.

### DIFF
--- a/consensus.go
+++ b/consensus.go
@@ -756,7 +756,7 @@ func (p *Pbft) GetProposal() *Proposal {
 	return p.state.proposal
 }
 
-// GetCurrentView returnes current view
+// GetCurrentView returns current view
 func (p *Pbft) GetCurrentView() *View {
 	return p.state.view.Copy()
 }
--- a/consensus.go
+++ b/consensus.go
@@ -756,11 +756,16 @@ func (p *Pbft) GetProposal() *Proposal {
 	return p.state.proposal
 }
 
+// GetCurrentView returnes current view
+func (p *Pbft) GetCurrentView() *View {
+	return p.state.view.Copy()
+}
+
 // getNextMessage reads a new message from the message queue
 func (p *Pbft) getNextMessage(span trace.Span, timeout time.Duration) (*MessageReq, bool) {
-	timeoutCh := time.After(timeout)
+	timeoutCh := p.notifier.CreateTimeoutChannel(timeout)
 	for {
-		msg, discards := p.notifier.ReadNextMessage(p)
+		msg, discards := p.notifier.ReadNextMessage(p, timeoutCh)
 		// send the discard messages
 		p.logger.Printf("[TRACE] Current state %s, number of prepared messages: %d, number of committed messages %d", PbftState(p.state.state), p.state.numPrepared(), p.state.numCommitted())
 
@@ -775,8 +780,9 @@ func (p *Pbft) getNextMessage(span trace.Span, timeout time.Duration) (*MessageR
 			return msg, true
 		}
 
-		// wait until there is a new message or
-		// someone closes the stopCh (i.e. timeout for round change)
+		// wait until there is a new message,
+		// someone closes the timeoutCh (i.e. timeout for round change) or
+		// cancel function is triggered
 		select {
 		case <-timeoutCh:
 			span.AddEvent("Timeout")
--- a/e2e/actions.go
+++ b/e2e/actions.go
@@ -1,12 +1,51 @@
 package e2e
 
 import (
+	"encoding/json"
 	"log"
 	"math/rand"
 	"time"
 
 	"github.com/0xPolygon/pbft-consensus"
 )
+
+const (
+	DropNode       string = "DropNode"
+	RevertDropNode string = "RevertDropNode"
+	Partition      string = "Partition"
+	LastSequence   string = "LastSequence"
+)
+
+// MetaData is a struct that holds data about fuzz actions that happened and need to be saved in .flow file
+type MetaData struct {
+	DataType string `json:"actionType"`
+	Data     string `json:"data"`
+	Sequence uint64 `json:"sequence"`
+	Round    uint64 `json:"round"`
+}
+
+// NewMetaData creates a new MetaData object
+func NewMetaData(dataType, data string, sequence, round uint64) *MetaData {
+	return &MetaData{
+		DataType: dataType,
+		Data:     data,
+		Sequence: sequence,
+		Round:    round,
+	}
+}
+
+// ConvertActionsToByteArrays converts ActionSerializable slice to JSON representation and returns it back as slice of byte arrays
+func ConvertActionsToByteArrays(actions []*MetaData) ([][]byte, error) {
+	var allRawACtions [][]byte
+	for _, message := range actions {
+		currentRawAction, err := json.Marshal(message)
+		if err != nil {
+			return allRawACtions, err
+		}
+		allRawACtions = append(allRawACtions, currentRawAction)
+	}
+	return allRawACtions, nil
+}
 
 type RevertFunc func()
 
@@ -35,10 +74,13 @@ func (dn *DropNodeAction) Apply(c *Cluster) RevertFunc {
 	log.Printf("Dropping node: '%s'.", nodeToStop)
 
 	c.StopNode(nodeToStop.name)
+	view := nodeToStop.GetCurrentView()
+	c.replayMessageNotifier.HandleAction(NewMetaData(DropNode, nodeToStop.name, view.Sequence, view.Round))
 
 	return func() {
 		log.Printf("Reverting stopped node %v\n", nodeToStop.name)
 		nodeToStop.Start()
+		c.replayMessageNotifier.HandleAction(NewMetaData(RevertDropNode, nodeToStop.name, c.GetMaxHeight()+1, 0))
 	}
 }
 
--- a/e2e/actions.go
+++ b/e2e/actions.go
@@ -1,7 +1,6 @@
 package e2e
 
 import (
-	"encoding/json"
 	"log"
 	"math/rand"
 	"time"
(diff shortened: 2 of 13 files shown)


e2e/fuzz/replay/replay_message_command.go
```diff
@@ -18,18 +17,20 @@ import (
...
+	metaDataFilePath string
 }
 
 // Help implements the cli.Command interface
-func (fc *ReplayMessageCommand) Help() string {
+func (rmc *ReplayMessageCommand) Help() string {
 	return `Runs the message and timeouts replay for analysis and testing purposes based on provided .flow file.
 	
-	Usage: replay-messages -file={fullPathToFlowFile}
+	Usage: replay-messages -messagesFile={fullPathToFlowFile} -metaDataFile={fullPathToFlowFile}
```
ferranbt, line 28: Is it required to have both? Could we merge both in the same file?

e2e/actions.go
```diff
@@ -1,13 +1,52 @@
...
 	"math/rand"
 	"time"
 
 	"github.com/0xPolygon/pbft-consensus"
 )
 
+const (
+	DropNode       string = "DropNode"
+	RevertDropNode string = "RevertDropNode"
+	Partition      string = "Partition"
```
ferranbt, line 15: Do we need this? As far as I remember, this is only required for node drop and we do not expect to do replay of other actions.

ferranbt (changes requested):
A lot of changes are required to support node drop. Do you guys have any insights on whether we can optimize any part?

e2e/fuzz/replay/replay_message_command.go
goran-ethernal: We could. Although its kinda messy to unmarshal the objects, since actions and messages are two different structures, and json.unmarshall is really dumb about these things, it needs to know the exact type in which to unmarshal, so it becomes a little bit tricky when reading the file.

goran-ethernal:
To answer your question, about optimizing. 80% of these changes are mostly just reorganizing code to different files and structures since it was all clustered up in the ReplayMessageNotifier, so we tried to separate logic for reading, writing and node execution to different files.

e2e/fuzz/replay/replay_message_command.go
goran-ethernal: ok, lets go then with the first option. Command will take folder of the two files as input

consensus.go
```diff
@@ -756,11 +756,16 @@ func (p *Pbft) GetProposal() *Proposal {
 	return p.state.proposal
 }
 
+// GetCurrentView returnes current view
```
Stefan-Ethernal, line 759: Small typo here:
```suggestion
// GetCurrentView returns current view
```

e2e/actions.go
```diff
@@ -1,13 +1,52 @@
...
+	return &MetaData{
+		DataType: dataType,
+		Data:     data,
+		Sequence: sequence,
+		Round:    round,
+	}
+}
+
+// ConvertActionsToByteArrays converts ActionSerializable slice to JSON representation and returns it back as slice of byte arrays
+func ConvertActionsToByteArrays(actions []*MetaData) ([][]byte, error) {
```
Stefan-Ethernal, line 37: Consider making this a generic functions, which would work with the slice of "objects", since it is pretty much the same as `ConvertMessagesToByteArrays` (and that one could be removed in that case). Tbh not entirely sure it would work, but it is worth of trying.

```suggestion
func ConvertToByteArrays(items []interface{}) ([][]byte, error) {
```

e2e/fuzz/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Daemon
+Fuzz daemon represents a test runner for fuzz/randomized tests. It is a separate go module/process, which runs alongside the cluster of Pbft nodes.
+It runs a Go routine and a single cluster with predefined number of nodes and randomly picks some subset of predefined actions and applies those to the cluster.
+
+Supported actions that can be simulated in fuzz daemon are:
+1. Drop Node - simulates dropping of a node in cluster where a single node goes offline, and is not communicating with the rest of the network.
+2. Partitions - simulates grouping of nodes in seperate partitions, where each partition only communicates with its member nodes.
+3. Flow Map - simulates a message routing mechanism where some of the nodes within the cluster are fully connected, whereas some of them are partially connected (namely only to the subset of peer nodes).
```
Stefan-Ethernal, line 19: Maybe remove Flow Map, since it is not yet merged. WDYT?

e2e/fuzz/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Replay Messages
+Fuzz framework relies on a randomized set of predefined actions, which are applied and reverted on **PolyBFT** nodes cluster. Since its execution is non-deterministic and algorithm failure can be discovered at any time, it is of utmost importance to have trace of triggers which led to the failure state and possibility to replay those triggers on-demand arbitrary number of times. This feature enables in-depth analysis of failure.
+
+Replay takes the provided `messages.flow` and `metaData.flow` files, creates a cluster with appropriate amount of nodes with same name as in fuzz run, pushes all the messages in appropriate node queues and starts the cluster. This enables quick replay of previously run execution and a way to analyze the problem that occurred on demand.
+
+To run replay messages, run the following command:
+
+`go run ./e2e/fuzz/cmd/main.go replay-messages -messagesFile={fullPathToMessagesFile} -metaDataFile={fullPathToMetaDataFile}`
+
+e.g., `go run ./e2e/fuzz/cmd/main.go replay-messages -messagesFile=../SavedData/messages.flow -metaDataFile=../SavedData/metaData.flow`
```
Stefan-Ethernal: Don't forget to update usage example, if you are going to provide only single CLI flag, denoting folder location where messages and metadata files reside.

e2e/fuzz/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+**NOTE**: Replay does not save .flow files on its execution.
+
+## Known issues
+When saving messages that are gossiped during the execution of fuzz daemon, messages will be sorted by sequence but they will not be in the order that they were gossiped, since `Gossip` method sends messages to each node asynchronously in separate go routins for each receiver. This means that a `PrePrepare` message may not be first in the `.flow` file since all the other messages are also sent in seperate routines. This does not have an effect or causes an issue on replay, since messages are seperated in different queues in **PolyBFT** state machine depending on its message type (`PrePrepare`, `Prepare`, `Commit`, `RoundChange`).
```
Stefan-Ethernal, line 45:
```suggestion
When saving messages that are gossiped during the execution of fuzz daemon, messages will be sorted by sequence but they will not be in the order that they were gossiped, since `Gossip` method sends messages to each node asynchronously in separate go routins for each receiver. This means that a `PrePrepare` message may not be first in the `.flow` file since all the other messages are also sent in seperate routines. This does not have an effect or causes an issue on replay, since messages are seperated in different queues in **PBFT** state machine depending on its message type (`PrePrepare`, `Prepare`, `Commit`, `RoundChange`).
```

e2e/fuzz/replay/replay_message.go
```diff
@@ -31,8 +31,8 @@ func NewReplayTimeoutMessage(to pbft.NodeID, msgType pbft.MsgType, view *pbft.Vi
 	}
 }
 
-// ConvertToByteArrays converts ReplayMessage slice to JSON representation and return it back as slice of byte arrays
-func ConvertToByteArrays(messages []*ReplayMessage) ([][]byte, error) {
+// ConvertMessagesToByteArrays converts ReplayMessage slice to JSON representation and return it back as slice of byte arrays
+func ConvertMessagesToByteArrays(messages []*ReplayMessage) ([][]byte, error) {
```
Stefan-Ethernal, line 35: This one could be removed if it turns out that generic flavor of fore-mentioned function does its job.

e2e/fuzz/replay/replay_message_reader.go
```diff
@@ -101,11 +138,9 @@ LOOP:
 			for _, message := range messages {
 				node, exists := nodes[string(message.To)]
 				if !exists {
-					log.Printf("[WARNING] Could not find node: %v to push message from .flow file.\n", message.To)
+					log.Printf("[WARNING] Could not find node: %v to push message from .flow file", message.To)
```
Stefan-Ethernal, line 149: We don't need new line here? Is it due to the fact that logger itself adds new line at the end, each time it gets invoked?

e2e/actions.go
```diff
@@ -1,13 +1,52 @@
...
 
+const (
+	DropNode       string = "DropNode"
+	RevertDropNode string = "RevertDropNode"
+	Partition      string = "Partition"
+	LastSequence   string = "LastSequence"
+)
+
+// MetaData is a struct that holds data about fuzz actions that happened and need to be saved in .flow file
+type MetaData struct {
```
Stefan-Ethernal, line 19: Could we move metadata into `replay_message.go`, since it is plain DTO?

Stefan-Ethernal (changes requested):

e2e/fuzz/README.md
Stefan-Ethernal: Marking this as resolved, since you have changed it in a mean time.

Stefan-Ethernal (approved):

PR #50, opened 2022-04-16
Use timer only at the round level

ferranbt:


state.go
```diff
@@ -210,13 +210,20 @@ type currentState struct {
...
 	err error
 }
 
 // newState creates a new state with reset round messages
 func newState() *currentState {
-	c := &currentState{}
+	c := &currentState{
+		// this is a default value, it will get reset
+		// at every iteration
+		timeout: time.NewTimer(0),
```
0xSasaPrsic, line 225: I think we should have default timeout greater than 0 so that there is a chance to read message from the queue for the round 0.

0xSasaPrsic, line 225: I think we should have default timeout greater than 0 so that there is a chance to read message from the queue for the round 0.

0xSasaPrsic (approved):

Stefan-Ethernal (approved):

PR #53, opened 2022-05-12
POS3-141 Sealing with node id

igorcrevar:
We need to expand `SealedProposal` => `CommittedSeals` to have node id, not just signature bytes

### DIFF
--- a/state.go
+++ b/state.go
@@ -181,6 +181,14 @@ func (p *Proposal) Copy() *Proposal {
 	return pp
 }
 
+type CommitedSeal struct {
+	// Signature value
+	Signature []byte
+
+	// Node that signed
+	NodeID NodeID
+}
+
 // currentState defines the current state object in PBFT
 type currentState struct {
 	// validators represent the current validator set
@@ -237,10 +245,10 @@ func (c *currentState) GetSequence() uint64 {
 	return c.view.Sequence
 }
 
-func (c *currentState) getCommittedSeals() [][]byte {
-	committedSeals := [][]byte{}
-	for _, commit := range c.committed {
-		committedSeals = append(committedSeals, commit.Seal)
+func (c *currentState) getCommittedSeals() []CommitedSeal {
+	var committedSeals []CommitedSeal
+	for nodeId, commit := range c.committed {
+		committedSeals = append(committedSeals, CommitedSeal{Signature: commit.Seal, NodeID: nodeId})
 	}
 	return committedSeals
 }
--- a/state.go
+++ b/state.go
@@ -181,7 +181,7 @@ func (p *Proposal) Copy() *Proposal {
 	return pp
 }
 
-type CommitedSeal struct {
+type CommittedSeal struct {
 	// Signature value
 	Signature []byte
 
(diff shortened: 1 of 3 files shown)


state.go
```diff
@@ -181,6 +181,14 @@ func (p *Proposal) Copy() *Proposal {
 	return pp
 }
 
+type CommitedSeal struct {
```
0xSasaPrsic, line 184: typo

state_test.go
```diff
@@ -310,35 +310,23 @@ func TestState_GetSequence(t *testing.T) {
...
-	foundSeal := false
-	for _, seal := range committedSeals {
-		if reflect.DeepEqual(seal, s.committed["A"].Seal) {
-			foundSeal = true
-		}
+	assert.Len(t, committedSeals, 3)
+	for _, commSeal := range committedSeals {
+		el := s.committed[commSeal.NodeID]
+		assert.NotNil(t, el)
+		assert.True(t, reflect.DeepEqual(commSeal.Signature, el.Seal))
```
0xSasaPrsic, line 327: bytes.Equal?

igorcrevar: fixed

0xSasaPrsic (dismissed):
Looks good.

state.go
```diff
@@ -181,6 +181,14 @@ func (p *Proposal) Copy() *Proposal {
 	return pp
 }
 
+type CommittedSeal struct {
+	// Signature value
+	Signature []byte
+
+	// Node that signed
+	NodeID NodeID
```
b00ris, line 189: Why do we need NodeID there?

state_test.go
```diff
@@ -310,35 +310,26 @@ func TestState_GetSequence(t *testing.T) {
...
 
-	assert.Len(t, committedSeals, 2)
-	assert.True(t, &committedSeals[0] != &s.committed["A"].Seal)
-	foundSeal := false
-	for _, seal := range committedSeals {
-		if reflect.DeepEqual(seal, s.committed["A"].Seal) {
-			foundSeal = true
-		}
+	assert.Len(t, committedSeals, 3)
+	processed := map[NodeID]bool{}
```
b00ris, line 324: The better way is to use map[NodeID]struct{}

state.go
Nemanja0x: We talked today during the sync that we will add NodeId so that we can use bls aggregated signatures

state_test.go
igorcrevar: i think it is easier this way otherwise we need to use this code in actual test
```
_, exists := processed[commSeal.NodeID]
assert.False(t, exists) 
```
But if you think it is necessary we will change it

b00ris: No, just put my thougts.

b00ris (dismissed):

state_test.go
ferranbt: I agree with @b00ris. `bool` does not represent anything there and unless semantically necessary, I prefer to use `struct{}`

ferranbt (dismissed):

state_test.go
igorcrevar: changed

PR #54, opened 2022-05-16
Enable race check for unit tests

b00ris:
Unit tests are run with race and shuffle flags enabled. Also getting and setting round to the `currentState` is now thread-safe.


Stefan-Ethernal (approved):
LGTM

PR #55, opened 2022-05-16
Add linters check

b00ris:


0xSasaPrsic (approved):
Nice one!

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,45 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
```
Stefan-Ethernal, line 7:
```suggestion
```
This may be removed, since we don't have master branch, but only `main`?

Stefan-Ethernal:
> Nice one!

Agree with @0xSasaPrsic. Good work! :+1: 
Have to notice that there weren't too much of violations either, so good work for us as well :partying_face: :wink:

Stefan-Ethernal (approved):

PR #57, opened 2022-05-24
Property based test

b00ris:
Closes https://linear.app/matic/issue/POS3-160/property-based-test-experiment-for-pbft-consensus


consensus.go
```diff
@@ -221,11 +244,7 @@ func (p *Pbft) SetBackend(backend Backend) error {
 
 // start starts the PBFT consensus state machine
 func (p *Pbft) Run(ctx context.Context) {
-	p.ctx = ctx
-
-	// the iteration always starts with the AcceptState.
-	// AcceptState stages will reset the rest of the message queues.
-	p.setState(AcceptState)
+	p.RunPrepare(ctx)
```
0xSasaPrsic, line 247: Since this method is making initial state of the state machine, maybe we can rename it?

consensus.go
```diff
@@ -244,6 +263,17 @@ func (p *Pbft) Run(ctx context.Context) {
 	}
 }
 
+func (p *Pbft) RunPrepare(ctx context.Context) {
```
0xSasaPrsic, line 266: Since this method is making initial state of the state machine, maybe we can rename it?

state.go
```diff
@@ -238,7 +238,7 @@ func newState() *currentState {
 }
 
 func (c *currentState) IsLocked() bool {
-	return c.locked
+	return atomic.LoadUint64(&c.locked) == 1
```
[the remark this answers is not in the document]
b00ris: I'm not changing value. Just taking the address.
```func TestAtomic(t *testing.T) {
	var v uint64
	ch := make(chan struct{})
	for i := 0; i < 5; i++ {
		i := i
		go func() {
			fmt.Printf("%d %p\n", i, &v)
			for {
				select {
				case <-ch:
					return
				default:
					atomic.StoreUint64(&v, atomic.LoadUint64(&v)+1)
				}
			}
		}()
	}
	time.Sleep(time.Second * 2)
	close(ch)
	fmt.Println(atomic.LoadUint64(&v))
}```

File: consensus.go [hunk 4]
```diff
@@ -858,3 +886,13 @@ func MaxFaultyNodes(nodesCount int) int {
 func QuorumSize(nodesCount int) int {
 	return 2*MaxFaultyNodes(nodesCount) + 1
 }
+
+//Line - returns line of code where it is called
+func Line() string {
```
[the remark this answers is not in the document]
b00ris: Just a debug code. Removed

e2e/rapid_test.go
```diff
@@ -0,0 +1,579 @@
...
+			}
+			if callNumber > 100 {
+				t.Error("stucked")
+				return
+			}
+		}
+	})
+}
+
+func TestCheckLivenessIssueCheck_Fails(t *testing.T) {
```
ferranbt, line 218: Prefix the test.

e2e/rapid_test.go
```diff
@@ -0,0 +1,579 @@
...
+			if getMaxClusterRound(cluster) > 3 {
+				t.Error("Infinite rounds")
+				return
+			}
+		}
+	})
+
+}
+
+func TestSeveralNodesCanSwitchToDoneState(t *testing.T) {
```
ferranbt, line 142: Prefix the test

e2e/rapid_test.go
```diff
@@ -0,0 +1,579 @@
...
+			0: {0, 1, 3, 2},
+			1: {0, 1, 3},
+			2: {0, 1, 2, 3},
+		},
+	}
+
+	countPrepare := 0
+	ft := &fakeTransport{
+		GossipFunc: func(ft *fakeTransport, msg *pbft.MessageReq) error {
+			routing, changed := rounds[msg.View.Round]
```
ferranbt, line 231: It would be good to have an explanation of what is this transport checking.

e2e/rapid_test.go
```diff
@@ -0,0 +1,579 @@
...
+	}
+	cluster, timeoutsChan := generateCluster(numOfNodes, ft)
+	for i := range cluster {
+		cluster[i].SetInitialState(context.Background())
+	}
+
+	stuckList := NewBoolSlice(len(cluster))
+	doneList := NewBoolSlice(len(cluster))
+
+	callNumber := 0
```
ferranbt, line 349: Is there a way to abstract all this process? It seems to me that it is always doing the same thing:
- Get `stuck` and `done` list
- For loop over call number.
- Call `runClusterCycle`
- `setDoneOnDoneState` and do `timeoutsChan`.
- Check `doneList`.
- Check `getMaxClusterRound`.

ferranbt (changes requested):

consensus.go
```diff
@@ -244,6 +262,17 @@ func (p *Pbft) Run(ctx context.Context) {
 	}
 }
 
+func (p *Pbft) SetInitialState(ctx context.Context) {
```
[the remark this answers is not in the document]
b00ris: Just extracted to a separate function to have a bit more cleaner tests.

e2e/rapid_test.go
```diff
@@ -0,0 +1,579 @@
...
+	"io"
+	"log"
+	"pgregory.net/rapid"
+	"strconv"
+	"sync"
+	"testing"
+	"time"
+)
+
+const waitDuration = 50 * time.Millisecond
```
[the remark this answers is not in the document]
b00ris: I think you are right.  I suggest to create a separate issue for it, because I'm using some structures from e2e.

consensus.go
```diff
@@ -178,6 +179,10 @@ type Pbft struct {
 
 	// notifier is a reference to the struct which encapsulates handling messages and timeouts
 	notifier StateNotifier
+
+	//for experimenting
+	roundTimeoutFunc  func() <-chan time.Time
+	proposalDelayFunc func(ctx context.Context, delay time.Duration)
```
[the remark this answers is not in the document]
b00ris: Just to simplify tests. I don't need to think about proposal time for writing tests.

.gitignore
```diff
@@ -3,4 +3,5 @@ SavedState/
 **/*.flow
 **/*.log
 **/*.txt
-log*/
\ No newline at end of file
+log*/
+e2e/*.fail
```
Stefan-Ethernal, line 7: Seems that this is a leftover, since I don't see we are generating .fail files in the e2e folder.

e2e/rapid_test.go
```diff
@@ -0,0 +1,598 @@
...
+			case <-exitCh:
+			case <-deadline:
+			}
+
+			//useful for debug
+			_, _, _ = state, wgTime, isLocked
+			//if time.Since(wgTime) > waitDuration {
+			//	fmt.Println("wgitme ", state, i, callNumber, time.Since(wgTime), err1, isLocked)
+			//}
+			//
```
Stefan-Ethernal, line 485: Is this part necessary to leave or is it a leftover?

e2e/rapid_test.go
```diff
@@ -0,0 +1,598 @@
...
+			return ctx.Err()
+		default:
+
+		}
+	}
+}
+
+//generateRoundsRoutingMap generator for routing map
+//map[uint64]map[uint64][]uint64
+func generateRoundsRoutingMap(numOfNodes uint64) *rapid.Generator {
```
Stefan-Ethernal, line 570: Seems like this function is currently not used, so maybe consider removing it?

e2e/rapid_test.go
```diff
@@ -0,0 +1,598 @@
...
+		rapid.Uint64Range(0, 3),
+		//validator routing
+		generateRoutingMap(numOfNodes, 0),
+		//min num of rounds
+		1,
+		//max num of rounds
+		10)
+}
+
+func generateRoutingMap(numberOfNode uint64, minNumberOfConnections int) *rapid.Generator {
```
Stefan-Ethernal, line 582: If we are going to remove `generateRoundsRoutingMap`, we don't need this one either.

consensus.go
```diff
@@ -178,6 +179,10 @@ type Pbft struct {
 
 	// notifier is a reference to the struct which encapsulates handling messages and timeouts
 	notifier StateNotifier
+
+	//for experimenting
+	roundTimeoutFunc  func() <-chan time.Time
```
Stefan-Ethernal, line 184: Could you somehow adapt and use this `RoundTimeout` callback? It's intention was to set custom delegate which holds custom logic for round timeouts calculation? I see that you are returning a channel here, therefore there might be needed some adjustments on that callback, but if `roundTimeoutFunc` is meant for timeout mocking, then I would say that we already have support for that via `RoundTimeout` callback.

```Go
type Config struct {
   ...
   // RoundTimeout is a function that calculates timeout based on a round number
   RoundTimeout RoundTimeout
   ...
}
```

e2e/rapid_test.go
```diff
@@ -0,0 +1,598 @@
...
+	"pgregory.net/rapid"
+	"strconv"
+	"sync"
+	"testing"
+	"time"
+)
+
+const waitDuration = 50 * time.Millisecond
+
+type fakeTransport struct {
```
Stefan-Ethernal, line 21: Could we maybe reuse some of the existing implementation in e2e framework? We have similar implementation with `genericGossipTransport` (with it you are able to attach custom callback which can be used as a filter for messages delivery).

e2e/rapid_test.go
```diff
@@ -0,0 +1,598 @@
...
+
+	for _, node := range ft.nodes {
+		if msg.From != node.GetValidatorId() {
+			node.PushMessage(msg.Copy())
+		}
+	}
+	return nil
+}
+
+type BackendFake struct {
```
Stefan-Ethernal, line 39: Could we reuse some of the e2e backend logic here as well (embedding it and overriding the functions which aren't generic enough for property based testing framework)?

Stefan-Ethernal (changes requested):

e2e/rapid_test.go
```diff
@@ -0,0 +1,598 @@
...
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+		defer cancel()
+
+		err := runCluster(ctx,
+			cluster,
+			sendTimeoutIfNNodesStucked(timeoutsChan, numOfNodes),
+			func(doneList *BoolSlice) bool {
+				//check that 3 node switched to done state
```
0xSasaPrsic, line 115: I don't think this comment is correct.

consensus.go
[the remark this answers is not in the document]
ferranbt: How does this help for testing?

consensus.go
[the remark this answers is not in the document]
ferranbt: But could we get the same result just by setting the delay in the block proposal? Then, I would be on favour of using that instead of another test wrapper.

.gitignore
b00ris: When you property based test fails it generates *.fail file that you can use to reproduce the case.

consensus.go
ferranbt: Agree with @Stefan-Ethernal. The idea of these wrappers was to remove externalities on the state machine. I would like to avoid wrappers that build on top of other wrappers.

ferranbt (changes requested):

consensus.go
```diff
@@ -195,18 +200,35 @@ func New(validator SignKey, transport Transport, opts ...ConfigOption) *Pbft {
 		state:        newState(),
 		transport:    transport,
 		msgQueue:     newMsgQueue(),
-		updateCh:     make(chan struct{}),
+		updateCh:     make(chan struct{}, 100), //hack. There is a bug when you have several messages pushed on the same time.
```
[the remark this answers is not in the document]
b00ris: We use in PushInternal such code:

```code	

select {

case p.updateCh <- struct{}{}:

default:

}```

that notifies getNextMessage that we have a new message. 

If we haven't read the previous message from p.updateCh we'll fall to default and won't get an update.  I've just added a buffer to make it not reproducible. It may be not a problem from v3(Need to be investigated).

File: e2e/rapid_test.go (hunk 13 above)
Comment (b00ris) [reply e2e/rapid_test.go:485]: I'd prefer to save it for debug purposes.

File: e2e/rapid_test.go (hunk 14 above)
Comment (b00ris) [reply e2e/rapid_test.go:570]: We need routing map generation from rapid for generating minimal reproducible case.

Comment (Stefan-Ethernal) [reply e2e/rapid_test.go:570]: Ok, but my point was that I don't see it is called anywhere explicitly (neither its name is passed to the rapid framework). Am I missing out something?

File: e2e/rapid_test.go (hunk 17 above)
Comment (b00ris) [reply e2e/rapid_test.go:21]: I thought about it. I didn't manage to do it simply. So I created a separate issue for it https://linear.app/matic/issue/POS3-184/property-tests-toolkit-refactoring

File: e2e/rapid_test.go (hunk 14 above)
Comment (b00ris) [reply e2e/rapid_test.go:570]: I see. You are right. I removed a test that used that function.

File: e2e/rapid_test.go (hunk 18 above)
Comment (b00ris) [reply e2e/rapid_test.go:39]: I suggest to resolve it as a part of https://linear.app/matic/issue/POS3-184/property-tests-toolkit-refactoring

File: consensus.go (hunk 16 above)
Comment (b00ris) [reply consensus.go]: I like your points! It became much simpler!

File: state.go [hunk 21]
```diff
@@ -216,10 +216,10 @@ type currentState struct {
 	roundMessages map[uint64]map[NodeID]*MessageReq
 
 	// Locked signals whether the proposal is locked
-	locked bool
+	locked uint64
```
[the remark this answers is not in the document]
b00ris: I used the atomic package for changing `locked` variable because it's used in set/getState also(just to be more identical). 
But I agree It may be a possible solution to use sync.Mutex/RWMutex there. Do you see how mutexes make the code better?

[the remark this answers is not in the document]
vcastellm: Basically we won't need to maintain our own mutex so code and tests will be reduced, less code == better code :)

vcastellm (dismissed):
LGTM overall

consensus.go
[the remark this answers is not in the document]
Stefan-Ethernal: @b00ris not sure have you seen explanation for such pattern on task you have created (https://linear.app/matic/issue/POS3-183/missing-update).
I'm going to paste explanation here as well, just to have conversation streamlined:

We have previously discussed with @ferranbt about part of code used here and were asking the same question. 🙂 

However, as Ferran explained, the thing is that `updateCh` is used just as a trigger, which signals to start next iteration of messages reading loop in the `getNextMessage` function. Therefore we don't need to act upon each message being pushed into the queue, as we are going to read it eventually (or discard if we accumulate enough messages for the given round and sequence to proceed to the next Pbft state).

consensus.go
```diff
@@ -29,7 +30,7 @@ type Config struct {
 	// Tracer is the OpenTelemetry tracer to log traces
 	Tracer trace.Tracer
 
-	// RoundTimeout is a function that calculates timeout based on a round number
+	// RoundTimeoutFunction is a function that calculates timeout based on a round number
```
Stefan-Ethernal, line 33:
```suggestion
	// RoundTimeout is a function that calculates timeout based on a round number
```

e2e/rapid_test.go
```diff
@@ -0,0 +1,573 @@
...
+				return false
+			}, 100)
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
+}
+
+func TestProperty4NodesCanAchiveAgreementIfWeLockButNotCommitProposer_Fails(t *testing.T) {
+	t.Skip("Unskip when fix")
```
Stefan-Ethernal, line 200: This is skipped due to the issue on the consensus algorithm (one of the liveness issues I believe; looks like the one where there is not quorum of commit messages in order to insert a proposal)? Or the test needs to be re-worked?

consensus.go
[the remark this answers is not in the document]
ferranbt: This would be my only minor concern here. It is okay if we refactor the rest of the property-based toolkit as another PR but I am not sure this behaves correctly as it is right now with the buffer.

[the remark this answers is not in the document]
b00ris: @Stefan-Ethernal @ferranbt I've checked it again. Tests become probabilistic if i remove capacity.
I'm not sure that I got your idea. As i understand updateCh used as a trigger to process one more message. There is a function that reads messages.
```
func (p *Pbft) getNextMessage(span trace.Span) (*MessageReq, bool) {
	for {
		msg, discards := p.notifier.ReadNextMessage(p)
...
		// wait until there is a new message or
		// someone closes the stopCh (i.e. timeout for round change)
		select {
		case <-p.state.timeoutChan:
...
		case <-p.ctx.Done():
			return nil, false
		case <-p.updateCh:
		}
}
```
For example: 
Cluster ABCD. A should receive 3 Prepare messages.
So we need to call PushMessage for B, C, D. It adds a message to a queue and trigger updateCh.
```
func (p *Pbft) PushMessageInternal(msg *MessageReq) {
	p.msgQueue.pushMessage(msg)

	select {
	case p.updateCh <- struct{}{}:
	default:// we miss trigger here
	}
}
```
A receives B, C and D at the same time(trigger once).
A read B and process, then read  trigger from updateCh, read C and got stuck because there is no trigger for the last message. 
We have to wait for timeout message to exit getNextMessage cycle.
Or we have to guarantee that pbft.PushMessage can't be called concurrent. Does my explanation correlate with the result of your discussion?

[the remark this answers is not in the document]
Stefan-Ethernal: But the thing here is that `getNextMessage` shouldn't proceed to `select-case` block if there is a message for the current `sequence` (block height) and `round`. Only if it don't find appropriate message it waits either for new message (`updateCh` notification), timeout or cancellation to happen in order to again try to read message for current `sequence` and `round`. I mean pbft algorithm shouldn't get stuck, since we are having timeouts as well and in that case, if timeout elapses, it will jump out from `getNextMessage`.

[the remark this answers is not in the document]
b00ris: Yes, messages are correct. One node could stuck if it receive the message after ReadNextMessage but before select. I think the validator will and wait before timeout or next message that push to updateCh or receive created block from another peer.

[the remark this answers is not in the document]
Stefan-Ethernal: > One node could stuck if it receive the message after ReadNextMessage but before select.

If message is valid (namely for the current `sequence` and `round`), it is jumped out of the `getNextMessage` immediately after it is read and execution won't event fall down to `select-case`. Am I missing out something here? 

Are you talking here about use case where message queue is either empty or contains only messages for some past sequences/rounds (i.e. only messages that are going to be discarded). In such case, I believe that we could miss out some message notification, because of that `default` statement in `PushMessageInternal` and in that case timeout would occur. However, message is still injected to the queue and there is chance that it could be read or discarded on some other occasion. Is my understanding correct?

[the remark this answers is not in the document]
b00ris: The main problem with this thing is: to push in a not buffered channel we need someone who read from it at the same time(or we'll switch to default section). The simplest solution is to make channel buffered.

ferranbt (approved):

Stefan-Ethernal (approved):

PR #59, opened 2022-06-03
POS3-145: Report metrics from PBFT

momosh-ethernal:


stats/stats_test.go
```diff
@@ -0,0 +1,32 @@
...
+	stats.IncrMsgCount(uint32(3))
+
+	// assert Round Changes
+	assert.Equal(t, uint64(1), stats.msgCount[uint32(0)])
+	// assert Prepepares
+	assert.Equal(t, uint64(2), stats.msgCount[uint32(1)])
+	// assert Commit Messages
+	assert.Equal(t, uint64(1), stats.msgCount[uint32(2)])
+	// assert Prepare Message
+	assert.Equal(t, uint64(3), stats.msgCount[uint32(3)])
```
0xSasaPrsic, line 31: could you add reset assertion?

momosh-ethernal: sure thing 👍

consensus.go
```diff
@@ -293,7 +318,10 @@ func (p *Pbft) setRound(round uint64) {
 func (p *Pbft) runAcceptState(ctx context.Context) { // start new round
 	_, span := p.tracer.Start(ctx, "AcceptState")
 	defer span.End()
-
+	startTime := time.Now()
```
ferranbt, line 321: Why do not put this at the beginning of `runCycle`? You would only need to do it once per all the states.

ferranbt (changes requested):

consensus.go
```diff
@@ -549,6 +581,8 @@ func (p *Pbft) runCommitState(ctx context.Context) {
 	} else {
 		// move to done state to finish the current iteration of the state machine
 		p.setState(DoneState)
+		// todo reset stats?
+		//	p.stats.Clear()
```
b00ris, line 585: Do we need it?

stats/stats.go
```diff
@@ -0,0 +1,74 @@
+package stats
+
+import (
+	"sync"
+	"time"
+)
+
+type Stats struct {
+	// todo atomic instead of lock?
+	lock *sync.Mutex
```
b00ris, line 10: Not sure that it's possible to use atomic for maps.
But RWMutex will be suitable there.

momosh-ethernal: RWMutex seems more convenient, but I'm not convinced we need it here.
RWMutex would come in handy here if we dealt with multiple readers and a few writers. Also using RWMutex comes with a performance hit.

More meaningful optimization would be that we give up on those maps and introduce fields for each message type in the Stats itself, and then use atomics.

But for now, I think we're ok

consensus.go
momosh-ethernal: Comments you mean?
Guess not

consensus.go
```diff
@@ -413,9 +446,13 @@ func (p *Pbft) runAcceptState(ctx context.Context) { // start new round
 // runValidateState implements the Validate state loop.
 //
 // The Validate state is rather simple - all nodes do in this state is read messages and add them to their local snapshot state
-func (p *Pbft) runValidateState(ctx context.Context) { // start new round
+func (p *Pbft) runValidateState(ctx context.Context, start time.Time) { // start new round
```
ferranbt, line 449: Why is this necessary?

ferranbt (changes requested):

consensus.go
momosh-ethernal: ok, will move state duration updates outside the state functions and keep previous signature

stats/stats.go
b00ris: I thought a bit and realized that it's better to use Mutex here. It can guarantee that stats is consistent.

b00ris (dismissed):

ferranbt (dismissed):

stats/stats.go
```diff
@@ -0,0 +1,74 @@
...
+
+	for msgType, duration := range s.stateDuration {
+		stats.stateDuration[msgType] = duration
+	}
+
+	return *stats
+}
+
+// TODO: reset?
+// TODO: stats per round?
```
Stefan-Ethernal, line 67: We can remove these comments I suppose?

consensus.go
```diff
@@ -294,6 +323,8 @@ func (p *Pbft) runAcceptState(ctx context.Context) { // start new round
 	_, span := p.tracer.Start(ctx, "AcceptState")
 	defer span.End()
 
+	// todo backend init?
+	p.stats.SetView(p.state.view.Sequence, p.state.view.Round)
```
Stefan-Ethernal, line 327: If this is staying here, we can remove this todo comment...?

stats/stats.go
```diff
@@ -0,0 +1,74 @@
...
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.msgCount[msgType]++
+}
+
+func (s *Stats) StateDuration(msgType string, t time.Time) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.stateDuration[msgType] = time.Since(t)
+}
```
Stefan-Ethernal, line 44:
```suggestion
func (s *Stats) StateDuration(state string, t time.Time) {
	s.lock.Lock()
	defer s.lock.Unlock()
	s.stateDuration[state] = time.Since(t)
}
```

Stefan-Ethernal (changes requested):

Stefan-Ethernal (approved):

PR #60, opened 2022-06-03
Pos3 172 double signing test

0xSasaPrsic:


e2e/rapid_test.go
```diff
@@ -45,8 +47,7 @@ func (bf *BackendFake) Validate(proposal *pbft.Proposal) error {
 }
 
 func (bf *BackendFake) Insert(p *pbft.SealedProposal) error {
-	//TODO implement me
-	return nil
+	return bf.finalProposals.Insert(*p)
```
b00ris, line 50: could you make it non-mandatory(if it's not null)?

b00ris (approved):

PR #63, opened 2022-06-14
Add code coverage.

0xSasaPrsic:


Stefan-Ethernal (approved):

goran-ethernal (approved):

b00ris (approved):

b00ris (commented):
I think we are able to add 90% threshold there.

0xSasaPrsic:
> I think we are able to add 90% threshold there.

Currently, there should be no decrease in coverage since threshold is 0% for drop of the parent which is >90%.

PR #64, opened 2022-07-06
Add voting power

b00ris:
closes https://0x.youtrack.cloud/issue/V3-14


consensus.go
```diff
@@ -674,15 +702,32 @@ func (p *Pbft) runRoundChangeState(ctx context.Context) {
...
+			}
+		} else {
+			totalVotingPower := totalVotingPower(p.config.VotingPower)
+			validVotingPower := QuorumSizeVP(totalVotingPower)
+			roundVotingPower := p.state.calculateMessagesVotingPower(p.state.roundMessages[msg.View.Round], p.config.VotingPower)
+			if roundVotingPower >= validVotingPower {
+				// start a new round inmediatly
+				p.state.SetCurrentRound(msg.View.Round)
+				p.setState(AcceptState)
+			} else if roundVotingPower > MaxFaultyVP(totalVotingPower)+1 {
```
Nemanja0x, line 725: This condition should probably be 
`else if roundVotingPower > MaxFaultyVP(totalVotingPower)`
or 
`else if roundVotingPower >= MaxFaultyVP(totalVotingPower)+1`
?

consensus.go
```diff
@@ -899,3 +944,22 @@ func MaxFaultyNodes(nodesCount int) int {
 func QuorumSize(nodesCount int) int {
 	return 2*MaxFaultyNodes(nodesCount) + 1
 }
+
+func totalVotingPower(mp map[NodeID]uint64) uint64 {
```
Nemanja0x, line 948: Maybe this can be TotalVotingPower so we can use it in V3 code ?? Just a suggestion.

consensus.go
```diff
@@ -501,17 +511,35 @@ func (p *Pbft) runValidateState(ctx context.Context) { // start new round
 			panic(fmt.Errorf("BUG: Unexpected message type: %s in %s", msg.Type, p.getState()))
 		}
 
-		if p.state.numPrepared() > p.state.NumValid() {
-			// we have received enough prepare messages
-			sendCommit(span)
-		}
+		if p.config.VotingPower == nil {
```
Nemanja0x, line 514: There is a possibility not to use VotingPower, and go the old way?

0xSasaPrsic: Should we add some info here so that we know that if not set default is used?

consensus.go
```diff
@@ -674,15 +702,32 @@ func (p *Pbft) runRoundChangeState(ctx context.Context) {
...
-		if num == p.state.NumValid() {
-			// start a new round inmediatly
-			p.state.SetCurrentRound(msg.View.Round)
-			p.setState(AcceptState)
-		} else if num == p.state.MaxFaultyNodes()+1 {
-			// weak certificate, try to catch up if our round number is smaller
-			if p.state.GetCurrentRound() < msg.View.Round {
-				// update timer
-				sendRoundChange(msg.View.Round)
+		if p.config.VotingPower == nil {
```
0xSasaPrsic, line 705: Should we add some info here so that we know that if not set default is used?

consensus.go
b00ris: The idea was to be able to select counting and voting power algorithms. I'm not sure that i understood your question @Nemanja0x.
@0xSasaPrsic Added comments

consensus.go
b00ris: Aggree. Done

consensus.go
b00ris: Right. Thank you!

consensus.go
Nemanja0x: "The idea was to be able to select counting and voting power algorithms" - this is the answer. :)

Nemanja0x (dismissed):

0xSasaPrsic (approved):

consensus.go
```diff
@@ -899,3 +950,22 @@ func MaxFaultyNodes(nodesCount int) int {
...
+}
+
+func MaxFaultyVP(vp uint64) uint64 {
+	if vp == 0 {
+		return 0
+	}
+	return (vp - 1) / 3
+}
+
+func QuorumSizeVP(totalVP uint64) uint64 {
```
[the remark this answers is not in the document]
b00ris: Discussed in a chat and decided to keep as is. We'll need to return back on integration phase.

Nemanja0x (approved):

PR #66, opened 2022-07-12
POS3-87: Source code organizational changes

begmaroman:
- Refactored structure of e2e
- Updated Makefile with a new command


Makefile
```diff
@@ -1,11 +1,16 @@
+nodes=5
+duration=25m
 
 test:
 	go test -v --race -shuffle=on -coverprofile=coverage.out -covermode=atomic ./...
 
 e2e:
```
Stefan-Ethernal, line 7: I would rename this to e.g. `fuzz-e2e`.

e2e/replay/msg_persister.go
```diff
@@ -44,25 +44,25 @@ func (r *replayMessagePersister) saveMetaData(nodeNames *[]string) error {
 	return err
 }
 
-// saveCachedMessages saves currently cached messages and timeouts to .flow file
-func (r *replayMessagePersister) saveCachedMessages() error {
+// SaveCachedMessages saves currently cached messages and timeouts to .flow file
+func (r *MessagePersister) SaveCachedMessages() error {
```
Stefan-Ethernal, line 48: Seems like this doesn't need to be public eventually (it is called by `MessageNotifier`, which is in the `replay` package as well)

e2e/replay/msg_persister.go
```diff
@@ -44,25 +44,25 @@ func (r *replayMessagePersister) saveMetaData(nodeNames *[]string) error {
...
+		err = r.SaveMessages(r.file)
 	}
 
 	return err
 }
 
-// createFile creates a .flow file to save messages and timeouts on the predifined location
-func (r *replayMessagePersister) createFile() error {
+// CreateFile creates a .flow file to save messages and timeouts on the predifined location
+func (r *MessagePersister) CreateFile() error {
```
Stefan-Ethernal, line 65: Seems like this doesn't need to be public eventually (it is called by `MessageNotifier`, which is in the `replay` package as well)

e2e/replay/msg_persister.go
```diff
@@ -86,23 +86,23 @@ func (r *replayMessagePersister) createFile() error {
 	return nil
 }
 
-// closeFile closes file created by the ReplayMessagesHandler if it is open
-func (r *replayMessagePersister) closeFile() error {
+// CloseFile closes file created by the ReplayMessagesHandler if it is open
+func (r *MessagePersister) CloseFile() error {
```
Stefan-Ethernal, line 90: Seems like this doesn't need to be public eventually (it is called by `MessageNotifier`, which is in the `replay` package as well)

e2e/replay/msg_persister.go
```diff
@@ -86,23 +86,23 @@ func (r *replayMessagePersister) createFile() error {
...
+func (r *MessagePersister) AddMessage(message *Message) {
 	r.lock.Lock()
 	defer r.lock.Unlock()
 	r.messages = append(r.messages, message)
 }
 
-// saveMessages saves ReplayMessages to the JSON file within the pre-defined directory.
-func (r *replayMessagePersister) saveMessages(fileWriter *os.File) error {
+// SaveMessages saves ReplayMessages to the JSON file within the pre-defined directory.
+func (r *MessagePersister) SaveMessages(fileWriter *os.File) error {
```
Stefan-Ethernal, line 105: Seems like this doesn't need to be public eventually (it is called by `MessageNotifier`, which is in the `replay` package as well)

e2e/replay/msg_persister.go
```diff
@@ -86,23 +86,23 @@ func (r *replayMessagePersister) createFile() error {
...
 	if r.file != nil {
 		return r.file.Close()
 	}
 	return nil
 }
 
-// addMessage adds a message from sequence to messages cache that will be written to .flow file
-func (r *replayMessagePersister) addMessage(message *ReplayMessage) {
+// AddMessage adds a message from sequence to message cache that will be written to .flow file
+func (r *MessagePersister) AddMessage(message *Message) {
```
Stefan-Ethernal, line 98: Seems like this doesn't need to be public eventually (it is called by `MessageNotifier`, which is in the `replay` package as well)

e2e/replay/msg_persister.go
```diff
@@ -12,17 +12,17 @@ import (
...
 	lock     sync.Mutex
-	messages []*ReplayMessage
+	messages []*Message
 	file     *os.File
 }
 
-// saveMetaData saves node meta data to .flow file
-func (r *replayMessagePersister) saveMetaData(nodeNames *[]string) error {
+// SaveMetaData saves node meta data to .flow file
+func (r *MessagePersister) SaveMetaData(nodeNames *[]string) error {
```
Stefan-Ethernal, line 23: Seems like this doesn't need to be public eventually (it is called in `MessageNotifier`, which is in the `replay` package as well)

e2e/replay/msg_middleware.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/0xPolygon/pbft-consensus"
+	"github.com/0xPolygon/pbft-consensus/e2e"
+)
+
+const (
+	FileName = "messages"
+)
+
+// MessagesNotifier is a struct that implements ReplayNotifier interface
+type MessagesNotifier struct {
```
Stefan-Ethernal, line 16: I was thinking about renaming this struct into `MessagesMiddleware`, as it represents more than just a "notifier". WDYT?

e2e/replay/msg.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// Message is a struct that represents a single json object in .flow file
+type Message struct {
+	To      pbft.NodeID      `json:"to"`
+	Message *pbft.MessageReq `json:"message"`
+}
+
+// NewMessageReq creates a new message to be written to .flow file
+func NewMessageReq(to pbft.NodeID, message *pbft.MessageReq) *Message {
```
Stefan-Ethernal, line 16: Rename this to `NewMessage`. (also consider making it private/package visibility)

Stefan-Ethernal (changes requested):
Nice one, much better organization than it was before. :partying_face: 

Although just consider comments (basically comments can be grouped into two categories):
- revisit all the functions and make public only ones which needs to be public,
- a couple of renames here and there.

Stefan-Ethernal:
Also @ferranbt had some ideas for this task as well, but I really cannot remember what it was, since it was a couple of months ago... :slightly_smiling_face:

Makefile
begmaroman: Done

e2e/replay/msg_persister.go
begmaroman: Done

e2e/replay/msg_persister.go
begmaroman: Done

e2e/replay/msg_persister.go
begmaroman: Done

e2e/replay/msg_persister.go
begmaroman: Done

e2e/replay/msg_middleware.go
begmaroman: Agree, done

e2e/action/drop_node.go
```diff
@@ -0,0 +1,36 @@
+package action
+
+import (
+	"log"
+	"math/rand"
+
+	"github.com/0xPolygon/pbft-consensus"
+	"github.com/0xPolygon/pbft-consensus/e2e"
```
b00ris, line 8: Do you think it's worth to move action to a separate package?
It seems we may face a cycle dependencies there.

Stefan-Ethernal: Personally, I find it cleaner this way. Actions are used exclusively for fuzz framework and therefore intention was to keep it in a separate package.

begmaroman: The app would not compile if there are cycle deps here. I think it looks better than having everything within one package/file. Made everything more structured. Let's say if you want to add a new action, you just need to create one more file in the actions package and it would be clear and understandable. 

Regarding deps in this case: main -> command -> actions -> e2e. Everything is ok with this.

Stefan-Ethernal (changes requested):
Just a small nitpick rename, see if you agree and we can move on as far as I'm concerned.

e2e/cmd/replay/command.go
```diff
@@ -4,25 +4,35 @@ import (
...
 	"github.com/0xPolygon/pbft-consensus"
 	"github.com/0xPolygon/pbft-consensus/e2e"
-	"github.com/mitchellh/cli"
+	"github.com/0xPolygon/pbft-consensus/e2e/replay"
 )
 
-// ReplayMessageCommand is a struct containing data for running replay-message command
-type ReplayMessageCommand struct {
+// MessageCommand is a struct containing data for running replay-message command
+type MessageCommand struct {
```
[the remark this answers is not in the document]
begmaroman: `replay` is the package name. Based on the most common go codestyle, you should avoid using the name of a pacakge in the types name. Like it is better to use `replay.Command` instead of `replay.ReplayCommand`. No needed to duplicate names. 

Updated to just `Command`.

Stefan-Ethernal (dismissed):
LGTM

e2e/cmd/replay/command.go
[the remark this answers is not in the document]
Stefan-Ethernal: Agreed.

e2e/cmd/fuzz/command.go
```diff
@@ -32,12 +39,12 @@ func (fc *FuzzCommand) Help() string {
 }
 
 // Synopsis implements the cli.Command interface
-func (fc *FuzzCommand) Synopsis() string {
+func (fc *Command) Synopsis() string {
```
vcastellm, line 42:
```suggestion
func (Command) Synopsis() string {
```

.github/workflows/pr.yml
```diff
@@ -39,4 +39,4 @@ jobs:
           go-version: '1.17'
       - name: Go fuzz test
         timeout-minutes: 20
-        run: make fuzz
\ No newline at end of file
+        run: make unit-fuzz
```
vcastellm, line 42:
```suggestion
        run: make unit-fuzz

```

e2e/cmd/fuzz/command.go
```diff
@@ -1,26 +1,33 @@
...
+// New is the constructor of Command
+func New(ui cli.Ui) *Command {
+	return &Command{
+		UI: ui,
+	}
+}
+
 // Help implements the cli.Command interface
-func (fc *FuzzCommand) Help() string {
+func (fc *Command) Help() string {
```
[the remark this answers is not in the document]
begmaroman: done

e2e/cmd/fuzz/command.go
```diff
@@ -32,12 +39,12 @@ func (fc *FuzzCommand) Help() string {
...
 
 // Synopsis implements the cli.Command interface
-func (fc *FuzzCommand) Synopsis() string {
+func (fc *Command) Synopsis() string {
 	return "Starts the PolyBFT fuzz runner"
 }
 
 // Run implements the cli.Command interface and runs the command
-func (fc *FuzzCommand) Run(args []string) int {
+func (fc *Command) Run(args []string) int {
```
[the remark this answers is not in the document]
begmaroman: done

e2e/cmd/fuzz/command.go
```diff
@@ -69,7 +75,7 @@ func (fc *FuzzCommand) Run(args []string) int {
 }
 
 // NewFlagSet implements the FuzzCLICommand interface and creates a new flag set for command arguments
-func (fc *FuzzCommand) NewFlagSet() *flag.FlagSet {
+func (fc *Command) NewFlagSet() *flag.FlagSet {
```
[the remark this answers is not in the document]
begmaroman: done

.github/workflows/pr.yml
begmaroman: done

ferranbt (approved):

vcastellm (approved):

Stefan-Ethernal (approved):

PR #67, opened 2022-07-14
POS3-184: Property tests toolkit refactoring

begmaroman:


begmaroman:
I haven't updated pbft (non-e2e) tests coz they are located within different go modules which makes it a bit tricky. I would like to move step-by-step instead of everything at once.

consensus_pbft.go
```diff
@@ -0,0 +1,785 @@
+package pbft
```
Stefan-Ethernal, line 1: Revert name of the file either to `consensus.go` or some other name if you can come up with some more appropriate. :slightly_smiling_face:

e2e/replay/msg_persister.go
```diff
@@ -126,3 +126,16 @@ func (r *messagePersister) saveMessages(fileWriter *os.File) error {
 	r.messages = nil
 	return nil
 }
+
+// convertToByteArrays converts ReplayMessage slice to JSON representation and return it back as slice of byte arrays
```
Stefan-Ethernal, line 130:
```suggestion
// convertToByteArrays converts message slice to JSON representation and return it back as slice of byte arrays
```

e2e/backend.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+// IntegrationBackend extends the pbft.Backend interface with additional e2e-related method
+type IntegrationBackend interface {
+	pbft.Backend
+	SetBackendData(n *node)
+}
+
+// BackendFake implements IntegrationBackend interface
+type BackendFake struct {
```
Stefan-Ethernal, line 19: Nice, this is essential change.

Stefan-Ethernal (changes requested):
Looks good, just a minor comments and considerations.

e2e/replay/msg_persister.go
begmaroman: fixed

e2e/helper/node.go
```diff
@@ -0,0 +1,47 @@
+package helper
+
+import "github.com/0xPolygon/pbft-consensus"
+
+type NodeID struct {
```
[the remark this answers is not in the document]
begmaroman: Done

consensus_pbft.go
begmaroman: Well, this file contains the pbft implementation of consensus. Do you think this is not the right name for the file?

consensus.go
```diff
@@ -950,22 +159,3 @@ func MaxFaultyNodes(nodesCount int) int {
 func QuorumSize(nodesCount int) int {
```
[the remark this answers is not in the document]
begmaroman: Makes sense, done

consensus.go
```diff
@@ -1,69 +1,26 @@
 package pbft
```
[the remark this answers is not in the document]
begmaroman: This file contains the consensus configuration setup. Just renamed it to `consensus_cfg.go` and moved those two functions to helper

consensus_pbft.go
Stefan-Ethernal: I was just thinking that `consensus.go` was already simple and good enough. Thought that you've renamed it this way just because you have reserved `consensus.go` for that file which you've now renamed to `consensus_cfg.go`.
I don't have problem to leave it like this if you find it better reflects the purpose. Tbh, it bothered me more naming that other file `consensus` than this naming `consensus_pbft`. 😄

e2e/helper/validator_set.go
```diff
@@ -0,0 +1,47 @@
+package helper
```
Stefan-Ethernal, line 1: Seems it is better to put this to `e2e` package itself. Opinions?

e2e/helper/helper.go
```diff
@@ -1,6 +1,7 @@
 package helper
```
Stefan-Ethernal, line 1: Renaming this package to utils maybe?
```suggestion
package utils
```

begmaroman: then we need to rename all helpers to utils :) This is more about the convention I think. If everyone agrees to use utils instead of helpers then it would make sense to rename it as well. Otherwise, we would jump from utils to helpers and vice versa

e2e/helper/validator_set.go
begmaroman: Makes sense, done

Stefan-Ethernal (approved):

e2e/helper/helper.go
b00ris: We can also use common for this)
I'll agree with both

b00ris (approved):

PR #68, opened 2022-07-21
[RFC-21] Set voting power in backend

b00ris:
Don't merge it. 
There is an interface changes. It must be merged with v3 changes


b00ris:
Merged in https://github.com/0xPolygon/pbft-consensus/pull/69

PR #69, opened 2022-07-22
Consolidate quorum and max faulty nodes calculations

Stefan-Ethernal:
Right now we have two implementations of `Quorum` and `MaxFaulty` functions:
`QuorumSize` and `QuorumSizeVP`,
`MaxFaultyNodes` and `MaxFaultyVP`.

This PR unifies these two implementations, by abstracting its logic to interface and two implementations (weighted vs non-weighted voting process).


consensus_pbft.go
```diff
@@ -74,9 +75,12 @@ type Pbft struct {
 	// Output logger
 	logger Logger
 
-	// Config is the configuration of the consensus
+	// config is the configuration of the consensus
 	config *Config
 
+	// votingMetadata encapsulates voting votingMetadata (such as quorum size calculation, max faulty nodes etc.)
+	votingMetadata VotingMetadata
```
ferranbt, line 82: This is not a parameter of `Pbft` but of the specific backend instance.

consensus_test.go
```diff
@@ -779,13 +779,13 @@ func TestRoundChange_PropertyMajorityOfVotingPowerAggreement(t *testing.T) {
 		validators := make(ValStringStub, numberOfNodes)
 		votingPower := make(map[NodeID]uint64)
 
-		var totalVotingPower uint64
 		for i := range validators {
-			validators[i] = NodeID(strconv.Itoa(i))
-			votingPower[validators[i]] = stake[i]
-			totalVotingPower += stake[i]
+			nodeId := NodeID(strconv.Itoa(i))
```
ferranbt, line 783: Let's not only do property-based testing here but also the traditional unit tests.

e2e/helper/helper.go
```diff
@@ -16,15 +17,17 @@ import (
 	"github.com/0xPolygon/pbft-consensus"
 )
 
+const trueString = "true"
```
momosh-ethernal, line 20: hol'up, do we really need this?

consensus_pbft.go
```diff
@@ -412,35 +420,38 @@ func (p *Pbft) runValidateState(ctx context.Context) { // start new round
...
 		}
 
-		//if voting power is disabled - count number of votes
-		if !p.IsVotingPowerEnabled() {
-			if p.state.numPrepared() > p.state.NumValid() {
-				// we have received enough prepare messages
-				sendCommit(span)
-			}
+		switch votingMetadata := p.votingMetadata.(type) {
+		case *NonWeightedVotingMetadata:
```
ferranbt, line 424: I am not a fan of this switch, I would like to merge both things on the same function. This is increasing the logic by a lot.

ferranbt (changes requested):

e2e/helper/helper.go
Stefan-Ethernal: Just because of annoying linter warnings. :roll_eyes:

consensus_test.go
Stefan-Ethernal: Agreed. Already have one for `MaxFaulty` calculation, I'm going to add the one for `QuorumSize` as well.

e2e/helper/helper.go
Stefan-Ethernal: Although this is non-substantial change, agreed with Momčilo on chat, not to revert this back, because of linter.

begmaroman (approved):
Apart from Ferran's comments, looks good to me

consensus_pbft.go
Stefan-Ethernal: Any ideas how to make this more better? I'm using such `switch-case` construction on two places: `runValidateState` and `runRoundChangeState`. I could make some function, which would have two callbacks as parameters: `weightedCallbackFn` and `nonWeightedCallbackFn`. Any better idea?

Stefan-Ethernal: Any ideas how to make this more better? I'm using such `switch-case` construction on two places: `runValidateState` and `runRoundChangeState`. I could make some function, which would have two callbacks as parameters: `weightedCallbackFn` and `nonWeightedCallbackFn`. Any better idea?

ferranbt: To me, both are the same problem. You have a given quorum weight for the sequence, and each node has a weight. In the voting power case, the weight is the staked value and the node weight is their stake. In the case of the one-peer-one-vote, the weight is the number of peers and the node weight is one. I think you could simplify this with two backend functions, one to get the quorum weight and another to query the node weight.

Stefan-Ethernal: Ok I see your point, makes sense, will try to abstract that part as well.

consensus_pbft.go
Stefan-Ethernal: In that case, `Backend` implementation (e.g. `fsm`) provides `VotingMetadata` object to the consensus algorithm is that correct?

voting_metadata.go
```diff
@@ -0,0 +1,49 @@
+package pbft
+
+type VotingMetadata interface {
+	// QuorumSize returns PBFT message count needed to perform a single PBFT state transition
+	QuorumSize() uint64
```
ferranbt, line 5: I think these methods belong to the `ValidatorSet` object. These are fields of the validator set.

ferranbt (changes requested):

voting_metadata.go
Stefan-Ethernal: On the other hand, think that `votingPower` map should be part of `ValidatorSet` object (as well) and that `VotingMetadata` should receive it from it.

consensus_pbft.go
```diff
@@ -602,35 +593,20 @@ func (p *Pbft) runRoundChangeState(ctx context.Context) {
...
-				// weak certificate, try to catch up if our round number is smaller
-				if p.state.GetCurrentRound() < msg.View.Round {
-					// update timer
-					sendRoundChange(msg.View.Round)
-				}
+		_ = p.state.AddRoundMessage(msg)
+
+		currentVotingPower := p.votingMetadata.CalculateVotingPower(p.state.roundMessages[msg.View.Round].extractNodeIds())
+		// Round change quorum is 2*F round change messages (F denotes max faulty voting power)
+		requiredVotingPower := 2 * p.votingMetadata.MaxFaultyVotingPower()
```
b00ris, line 600: We need p.QuorumSize there, don't we(if i'm not mistaken)?

e2e/backend.go
```diff
@@ -18,6 +18,7 @@ type IntegrationBackend interface {
 // BackendFake implements IntegrationBackend interface
 type BackendFake struct {
 	nodes           []string
+	votingPowerMap  map[pbft.NodeID]uint64
```
b00ris, line 21: You may need https://github.com/0xPolygon/pbft-consensus/pull/68/files changes. It seems we can close(if you take all of this changes)/merge it there.

voting_metadata.go
b00ris: I aggree with  @ferranbt's point. It'd be great to move VotingMetadata to ValidatorSet interface. Initially i was thinking about the same but  tried not to change ValidatorSet interface. 
If we agreed about changing it we may simplify an interface to 1-2 functions. 
Like `HasQuorum(senders) bool`
`HasMaxFaultyVotingPower(senders) bool`
We get ValidatorSet for every block and  it knows when we reached quorum.

b00ris (changes requested):

b00ris:
@Stefan-Ethernal Good job!

consensus_pbft.go
Stefan-Ethernal: It is as it was before: for some reason quorum for round change message is `2*F` and not `QuorumSize`. Therefore didn't want to change that on my own.

consensus_pbft.go
```diff
@@ -572,7 +562,8 @@ func (p *Pbft) runRoundChangeState(ctx context.Context) {
 	} else {
 		// otherwise, it is due to a timeout in any stage
 		// First, we try to sync up with any max round already available
-		if maxRound, ok := p.state.maxRound(); ok {
+		// F + 1 round change messages for given round, where F denotes MaxFaulty is expected, in order to fast-track to maxRound
+		if maxRound, ok := p.state.maxRound(p.votingMetadata.MaxFaultyVotingPower() + 1); ok {
```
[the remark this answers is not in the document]
Stefan-Ethernal: Any suggestions? If we agree that quorum of round change messages is `2*F`, then we can use such quorum here as well.

voting_metadata.go
Stefan-Ethernal: If I understand your proposal well, don't like it just because I really think that pbft consensus library needs to offer (at least default) implementation for functions contained in `VotingMetadata` interface. Because otherwise library won't be encapsulated and self contained (if we were up to allow consumer of library to implement it arbitrarily).

e2e/backend.go
Stefan-Ethernal: Eventually left this part, as I've talked with @Nemanja0x and seems like he don't need `VotingPower` exposed in the `ValidatorSet` interface. Also, as it is unneeded by consensus library, therefore won't expand it.

e2e/cluster.go
```diff
@@ -339,6 +339,24 @@ func (c *Cluster) startNode(name string) {
 	c.nodes[name].Start()
 }
 
+// MaxFaulty is a wrapper function which invokes MaxFaultyVotingPower on PBFT consensus instance of the first node in cluster
+func (c *Cluster) MaxFaulty() (uint64, error) {
+	nodes := c.getNodes()
+	if len(nodes) == 0 {
+		return 0, errors.New("unable to determine max faulty nodes: cluster is empty")
```
begmaroman, line 346: Let's pre-define this error on the package level

e2e/cluster.go
```diff
@@ -339,6 +339,24 @@ func (c *Cluster) startNode(name string) {
...
+	}
+	return nodes[0].pbft.MaxFaultyVotingPower(), nil
+}
+
+// QuorumSize is a wrapper function which invokes QuorumSize on PBFT consensus instance of the first node in cluster
+func (c *Cluster) QuorumSize() (uint64, error) {
+	nodes := c.getNodes()
+	if len(nodes) == 0 {
+		return 0, errors.New("unable to determine max faulty nodes: cluster is empty")
```
begmaroman, line 355: Let's pre-define this error on the package level

Stefan-Ethernal: Oh, I've just noticed that this error message isn't correct (copy paste error).
It should be like this:
```go
errors.New("unable to determine max faulty nodes, because cluster is empty")
```
Do you still find beneficial to extract it to some package level variable?

begmaroman (approved):

voting_metadata.go
```diff
@@ -0,0 +1,52 @@
+package pbft
+
+// NewVotingMetadata initializes instance of VotingMetadata
+func NewVotingMetadata(votingPower map[NodeID]uint64) *VotingMetadata {
```
ferranbt, line 4: This data structure got weirder now. We return constant values, that are set inside the struct so they are not parametrizable and `CalculateVotingPower` is fixed. To me, this `Metadata` is just a struct with constant values.

ferranbt (changes requested):

voting_metadata.go
```diff
@@ -0,0 +1,52 @@
...
+	v.maxFaultyVotingPower = v.calculateMaxFaultyVotingPower()
+	v.quorumSize = 2*v.MaxFaultyVotingPower() + 1
+	return v
+}
+
+// VotingMetadata holds weighted vote for each validator. Weighted vote is based on its voting power (e.g. stake amount).
+// Voting power is used as input to calculate:
+//  - QuorumSize of entire validator set, M
+//  - MaxFaultyVotingPower, which is maximum tollerable value of malicious voting power in order to have Byzantine fault tollerant property satisfied.
+type VotingMetadata struct {
```
ferranbt, line 15: You do not need this `VotingMetadata` anymore, if `ValidatorSet` returns the voting power, add two atomic functions that return the Quorum and MaxFaultyVotingPower whenever the sequence starts and move CalculateVotingPower to the state.

ferranbt (changes requested):

voting_metadata.go
Stefan-Ethernal: The main purpose of `VotingMetadata` is that it can be reused in v3 client (we are using `QuorumSize` and `CalculateVotingPower`). If I'd move all of these to `state.go`, then we wouldn't be able to reuse it on v3 codebase. Agreed?

test_helpers.go
```diff
@@ -64,3 +64,16 @@ func (v *ValStringStub) Includes(id NodeID) bool {
...
+
+func (v *ValStringStub) VotingPower() map[NodeID]uint64 {
+	return CreateEqualWeightValidatorsMap(*v)
+}
+
+// CreateEqualWeightValidatorsMap is a helper function which creates map with same weight for every validator id in the provided slice
+func CreateEqualWeightValidatorsMap(validatorIds []NodeID) map[NodeID]uint64 {
+	weightedValidators := make(map[NodeID]uint64, len(validatorIds))
+	for _, validatorId := range validatorIds {
+		weightedValidators[validatorId] = 1
```
b00ris, line 76: This change confuses me because this struct is using in TestRoundChange_PropertyMajorityOfVotingPowerAggreement. It seems we need correct voting power there.

Stefan-Ethernal: I don't get it, we don't use this function (neither `ValStringStub`) in this particular test, but instead setting voting metadata explicitly to the `currentState` reference:
https://github.com/0xPolygon/pbft-consensus/pull/69/files#diff-b2f021d366e25d5b9980cd37386d09c568352c340e3bfebaf58034bb1a7000f4R819.

Seems like there is no problem to me. Although agree that property tests could be even more consolidated.

voting_metadata.go
ferranbt: But you can expose a function instead of a data structure
```
func CalculateQuorumPower() (uint64, uint64) {
   ...
}
```
And `CalculateVotingPower` is just an aggregation over a map. There is no point in having v3 depend on an external function to do that.

Stefan-Ethernal: Not only that, but also using `QuorumSize` as well: https://github.com/maticnetwork/v3/commit/656b17e00f8bfc68d29ac3f35f6f5c748a72aa05#diff-297043a61d1a54ac38dc24c4a73a7de2fac85bd5567b58b1cfb47d8bdff663bdR369.
I know that these functions are pretty simple and straight-forward to implement, but not sure why one would have same code on multiple places, since it is easy to reuse it via this small and encapsulated struct?

However, don't want to debate around this anymore, think that we've lost too much time discussing some relatively non-important points and decisions made. Will try to accommodate this feedback as well tomorrow.

ferranbt (dismissed):

ferranbt (approved):

begmaroman (approved):

consensus_pbft.go
```diff
@@ -606,35 +596,20 @@ func (p *Pbft) runRoundChangeState(ctx context.Context) {
...
-				// weak certificate, try to catch up if our round number is smaller
-				if p.state.GetCurrentRound() < msg.View.Round {
-					// update timer
-					sendRoundChange(msg.View.Round)
-				}
+		_ = p.state.AddRoundMessage(msg)
+
+		currentVotingPower := p.state.CalculateVotingPower(p.state.roundMessages[msg.View.Round].extractNodeIds())
+		// Round change quorum is 2*F round change messages (F denotes max faulty voting power)
+		requiredVotingPower := 2 * p.state.MaxFaultyVotingPower()
```
momosh-ethernal, line 603: magic number?

Stefan-Ethernal: It is implemented like that in the original IBFT algorithm as well and isn't something I've invented. :slightly_smiling_face:

b00ris (approved):

PR #70, opened 2022-07-27
Expand RoundInfo struct with previous and current round

Stefan-Ethernal:
Proposer calculation on v3 client depends on priority queue. We need current and previous round information provided from pbft consensus in order to know how many round elapsed in a mean time in order to make proposer calculation deterministic.

### DIFF
--- a/state.go
+++ b/state.go
@@ -22,6 +22,9 @@ type state struct {
 	// Current view
 	view *View
 
+	// Previous round
+	previousRound uint64
+
 	// List of prepared messages
 	prepared map[NodeID]*MessageReq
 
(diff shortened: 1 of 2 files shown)


ferranbt (approved):

state.go
```diff
@@ -215,7 +218,12 @@ func (s *state) GetCurrentRound() uint64 {
 	return atomic.LoadUint64(&s.view.Round)
 }
 
+func (s *state) GetPreviousRound() uint64 {
+	return atomic.LoadUint64(&s.previousRound)
+}
+
 func (s *state) SetCurrentRound(round uint64) {
+	atomic.StoreUint64(&s.previousRound, s.view.Round)
```
ferranbt, line 226: Smart.

PR #71, opened 2022-07-30
Voting power accumulation optimization

Stefan-Ethernal:
Instead of calculating accumulated voting power each time a message is retrieved from the queue, it is now cached and updated on each message retrieval.


b00ris (approved):
Good job!

ferranbt (approved):
Love it.

PR #73, opened 2022-08-04
Implement additional different voting power unit tests

Stefan-Ethernal:
Implement additional unit tests on PBFT which simulates different voting power scenarios:
1. There is quorum size of accumulated voting power of PREPARE messages, but no COMMIT messages => state machine remains in the `ValidateState`.
2. There is 2*F round change messages for greater round => state machine is in `AcceptState` with given round.

### DIFF
--- a/consensus_test.go
+++ b/consensus_test.go
@@ -473,12 +473,14 @@ func TestTransition_ValidateState_MoveToCommitState(t *testing.T) {
 		m.runCycle(context.Background())
 
 		m.expect(expectResult{
-			sequence:    1,
-			state:       CommitState,
-			prepareMsgs: 3,
-			commitMsgs:  3, // Commit messages (A proposer sent commit via state machine loop, C and D sent commit via emit message)
-			locked:      true,
-			outgoing:    1, // A commit message
+			sequence:               1,
+			state:                  CommitState,
+			prepareMsgs:            3,
+			commitMsgs:             3, // Commit messages (A proposer sent commit via state machine loop, C and D sent commit via emit message)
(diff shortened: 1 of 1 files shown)


ferranbt (changes requested):
Please split this into two PRs. One for the test and another one for the framework abstraction.

vcastellm (dismissed):
LGTM

consensus_test.go
```diff
@@ -489,17 +508,52 @@ func TestTransition_ValidateState_MoveToCommitState(t *testing.T) {
...
+		m.setState(ValidateState)
+		m.emitMsg(createMessage(NodeID("B"), MessageReq_Commit, nil))
+		m.emitMsg(createMessage(NodeID("C"), MessageReq_Commit, nil))
+
+		m.runCycle(context.Background())
+
+		m.expect(expectResult{
+			sequence:   1,
+			state:      RoundChangeState,
+			commitMsgs: 2, // Commit messages (A proposer sent commit via state machine loop, B sent commit via emit message)
```
ferranbt, line 537: We should add another condition here to check the aggregated voting power as well.

ferranbt (changes requested):

consensus_test.go
Stefan-Ethernal: Agreed.

Stefan-Ethernal: Fixed https://github.com/0xPolygon/pbft-consensus/pull/73/commits/386a435ae2576f0cc14ccebbff9c5c9faf3d8502

ferranbt (approved):

PR #76, opened 2022-08-06
Rename open telemetry traces

Stefan-Ethernal:
There were some duplicate trace names (`ValidateState`) which was confusing. Now all the traces are named consistent and uniquely, so their name suggests its purpose.

Although maybe we should consider not having separate traces for message reading loops in `AcceptState`, `ValidateState` and `RoundChangeState`, but those traces to be integral part of given state trace?

### DIFF
--- a/consensus_pbft.go
+++ b/consensus_pbft.go
@@ -367,8 +367,11 @@ func (p *Pbft) runAcceptState(ctx context.Context) { // start new round
 // The Validate state is rather simple - all nodes do in this state is read messages and add them to their local snapshot state
 func (p *Pbft) runValidateState(ctx context.Context) { // start new round
 	_, span := p.tracer.Start(ctx, "ValidateState")
-	// set the attributes of this span once it is done
-	defer p.setSpanStateAttributesWithTerminate(span)
+	// set the attributes of this span once runValidateState is done
+	defer func() {
+		p.setStateSpanAttributes(span)
+		span.End()
+	}()
 
 	hasCommitted := false
 	sendCommit := func(span trace.Span) {
@@ -428,12 +431,6 @@ func (p *Pbft) runValidateState(ctx context.Context) { // start new round
 	}
 }
 
-// setSpanStateAttributesWithTerminate sets state attributes to given span and terminates it
-func (p *Pbft) setSpanStateAttributesWithTerminate(span trace.Span) {
-	p.setStateSpanAttributes(span)
-	span.End()
-}
-
 // spanAddEventMessage reports given message to both PBFT built-in statistics reporting mechanism and open telemetry
 func (p *Pbft) spanAddEventMessage(typ string, span trace.Span, msg *MessageReq) {
 	p.stats.IncrMsgCount(msg.Type.String(), p.state.validators.VotingPower()[msg.From])
@@ -459,6 +456,9 @@ func (p *Pbft) spanAddEventMessage(typ string, span trace.Span, msg *MessageReq)
 func (p *Pbft) setStateSpanAttributes(span trace.Span) {
 	attr := []attribute.KeyValue{}
 
+	// round
+	attr = append(attr, attribute.Int64("round", int64(p.state.view.Round)))
+
 	// number of commit messages
 	attr = append(attr, attribute.Int("committed", p.state.numCommitted()))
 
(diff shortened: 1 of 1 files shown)


consensus_pbft.go
```diff
@@ -741,9 +751,9 @@ func (p *Pbft) getNextMessage(span trace.Span) (*MessageReq, bool) {
 			p.spanAddEventMessage("dropMessage", span, msg)
 		}
 		if msg != nil {
+			p.logger.Printf("[TRACE] Received %s", msg)
```
ferranbt, line 754: Why did you move this?

consensus_pbft.go
```diff
@@ -387,19 +393,21 @@ func (p *Pbft) runValidateState(ctx context.Context) { // start new round
 	}
 
 	quorum := p.state.getQuorumSize()
+	i := 1
 	for p.getState() == ValidateState {
-		_, span := p.tracer.Start(ctx, "ValidateState")
+		_, loopSpan := p.tracer.Start(ctx, fmt.Sprintf("ValidateState_Read#%d", i))
```
ferranbt, line 398: To me, this counter should be metadata of the trace, not part of the name itself.

consensus_pbft.go
```diff
@@ -321,13 +321,19 @@ func (p *Pbft) runAcceptState(ctx context.Context) { // start new round
 	// We only need to wait here for one type of message, the Prepare message from the proposer.
 	// However, since we can receive bad Prepare messages we have to wait (or timeout) until
 	// we get the message from the correct proposer.
+	i := 1
```
ferranbt, line 324: Please rename this to `counter`.

consensus_pbft.go
```diff
@@ -581,9 +589,11 @@ func (p *Pbft) runRoundChangeState(ctx context.Context) {
 		}
 	}
 
+	i := 1
 	// create a timer for the round change
 	for p.getState() == RoundChangeState {
-		_, span := p.tracer.Start(ctx, "RoundChangeState")
+		_, span := p.tracer.Start(ctx, fmt.Sprintf("RoundChange_Read#%d", i))
```
ferranbt, line 595: This variable gets called `span` and `loopSpan`. I would prefer to call it `span` always since there are no other spans around.

ferranbt (changes requested):

consensus_pbft.go
Stefan-Ethernal: Just wanted to make it consistent with the for loop above: https://github.com/0xPolygon/pbft-consensus/pull/76/files#diff-b208c241c10c010125c6a1c4f43ac407c5fb7b363bab10a050e5681019875478R750-R751. There we are first logging and than reporting metadata to the tracer, so wanted to keep it the same in this block as well.

I can revert it back,  if it bothers you.

consensus_pbft.go
Stefan-Ethernal: Is it ok to add attribute to the span this way or you have different idea?
`span.SetAttributes(attribute.Int("loop.counter", counter))`

I mean maybe we don't even need this `counter` at all. Intention was just to differ span within for loop with this outer span for handling specific state.

consensus_pbft.go
```diff
@@ -435,11 +433,6 @@ func (p *Pbft) runValidateState(ctx context.Context) { // start new round
 			// change to commit state just to get out of the loop
 			p.setState(CommitState)
 		}
-
-		// set the attributes of this span once it is done
-		p.setStateSpanAttributes(span)
```
ferranbt, line 440: This might be the reason why there was an inter-span. Imagine that the system gets locked in one of the cycles, we would not get any traces at all because it is waiting to exit.

ferranbt (changes requested):

consensus_pbft.go
Stefan-Ethernal: Are you referring that `runXXXState` (`runValidateState`, `runRoundChangeState` etc.) execution could stuck? How is that even possible (at least with the current implementation)? Eventually a timeout would elapse and the execution of a  PBFT state handler would come to an end.
If PBFT consensus is stuck (or locked as you are referring) it still means that we would have transitions as such: `AcceptState` -> `RoundChangeState` -> `AcceptState` etc.

So do you still find useful leaving inter-spans or not?

vcastellm: Can you elaborate on this?

Stefan-Ethernal: > Can you elaborate on this?

My point was to keep only single span through entire PBFT state handler (instead of reporting spans every time a message is read from the queue). 
Both approaches have positive and negative aspects: 
- **creating a span for each message retrieval** (like it was before) 
   pros: we are sure that we wouldn't miss any information, if (for some strange reason), we get stuck any of the `runXXXState` function.
   cons: in order to track which messages are aggregated through given PBFT state, we would need to analyze 
   separately each span, which can be tedious, since we don't have reported all the accumulated messages at the single 
   place, but reporting it one by one.
- **creating a single span for entire PBFT state handler**
  pros: all the information are aggregated and easily consumed
  cons: Ferran's concern [here](https://github.com/0xPolygon/pbft-consensus/pull/76#discussion_r940043739). Although I   don't see that it is likely to happen (as we have timeouts as a mechanism to break execution if it lasts too long).
  
On the end, I'm good with both approaches, just need to make an agreement and I'll accommodate it.

vcastellm: Yes I like this approach @Stefan-Ethernal, I was asking @ferranbt to explain his concerns 😄 because as you said, I don't see any stopper

ferranbt: As discussed, lets change this except for RoundChange state.

begmaroman (approved):

vcastellm (approved):

ferranbt (approved):

PR #77, opened 2022-08-08
Simplify unit test framework by using `createMessage` function

Stefan-Ethernal:
Simplify unit testing framework a little bit (use an existing helper function for `MessageReq` objects creation).

### DIFF
--- a/consensus_pbft.go
+++ b/consensus_pbft.go
@@ -316,8 +316,8 @@ func (p *Pbft) runAcceptState(ctx context.Context) { // start new round
 	// we are NOT a proposer for this height/round. Then, we have to wait
 	// for a pre-prepare message from the proposer
 
-	// We only need to wait here for one type of message, the Prepare message from the proposer.
-	// However, since we can receive bad Prepare messages we have to wait (or timeout) until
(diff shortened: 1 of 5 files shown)


ferranbt (changes requested):
I do prefer to keep the explicit `MessageReq` object. It is explicit so you know at every time what are you sending and the tests are short enough that we do not require this type of optimization.

Stefan-Ethernal:
> I do prefer to keep the explicit `MessageReq` object. It is explicit so you know at every time what are you sending and the tests are short enough that we do not require this type of optimization.

Don't get your point. Personally I don't like having repeated code all over the place and though that you also don't prefer doing the same thing multiple ways. Currently we have 3 ways to achieve the same thing: 
1. `createMessage` function (it exists from before),
2. instantiating `MessageReq` object directly
3. we even have `mockQueueMsg` function (https://github.com/0xPolygon/pbft-consensus/pull/77/files#diff-353efbbf5c50447329b1d508abf3f0d44552fd09f810ba1364d67bf5d7ef7f08L9).

My intention is clear: let's unify all those into single function and remove all the extra lines of code we don't need basically.

ferranbt (approved):

PR #78, opened 2022-08-11
Validate commit messages in parallel (WIP)

Stefan-Ethernal:


consensus_pbft.go
```diff
@@ -290,6 +291,8 @@ func (p *Pbft) runAcceptState(ctx context.Context) { // start new round
 				p.setState(RoundChangeState)
 				return
 			}
+			// TODO: Maybe can be removed from here
+			p.msgQueue.processPendingCommitMessages(p.state.view.Copy(), p.verifyCommitMessage)
```
ferranbt, line 295: Could we move this to the beginning of `ValidateState`? alongside the one in L366.

msg_queue.go
```diff
@@ -28,6 +32,27 @@ func (m *msgQueue) pushMessage(message *MessageReq) {
...
+// processPendingCommitMessages processes accumulated pending commit messages for the given sequence and round.
+func (m *msgQueue) processPendingCommitMessages(view *View, handler func(*MessageReq)) {
+	for {
+		m.queueLock.Lock()
+		msg, _ := m.pendingCommitMsgsQueue.readMessageWithDiscardsLocked(view, ValidateState)
+		m.queueLock.Unlock()
+		if msg == nil {
+			return
+		}
+		go handler(msg)
```
ferranbt, line 52: What happens if:
- There are 10 items on the pending queue.
- After `AcceptState` they get unlocked and verified as part of `processPendingCommitMessages` (this is concurrent).
- The round changes and the old verification is still going. The proposal will be wrong.

I do not see this function being totally idempotent.

msg_queue.go
```diff
@@ -28,6 +32,27 @@ func (m *msgQueue) pushMessage(message *MessageReq) {
...
+// pushPendingCommitMessage adds a new commit message to a message queue
+func (m *msgQueue) pushPendingCommitMessage(message *MessageReq) {
+	m.queueLock.Lock()
+	defer m.queueLock.Unlock()
+
+	m.pendingCommitMsgsQueue.Push(message)
+}
+
+// processPendingCommitMessages processes accumulated pending commit messages for the given sequence and round.
+func (m *msgQueue) processPendingCommitMessages(view *View, handler func(*MessageReq)) {
```
ferranbt, line 44: Why not make this `handler` a field inside `msgQueue`? It is always a reference to the same function.

msg_queue.go
```diff
@@ -38,44 +63,8 @@ func (m *msgQueue) readMessageWithDiscards(st State, current *View) (*MessageReq
...
-		if cmpView(msg.View, current) < 0 {
-			// old value, try again
-			discarded = append(discarded, msg)
-			continue
-		}
-
-		// good value, return it
-		return msg, discarded
-	}
+	return queue.readMessageWithDiscardsLocked(current, st)
```
ferranbt, line 67: Is there any benefit on moving this logic out?

consensus_pbft.go
```diff
@@ -792,6 +793,31 @@ func (p *Pbft) PushMessage(msg *MessageReq) {
...
 	}
 
+	// wanted to do Commit Validation in parallel
+	// for that reason, validating logic was moved from runValidateState
+	if msg.Type == MessageReq_Commit {
+		p.msgQueue.pushPendingCommitMessage(msg)
+		if p.state.view == nil {
+			return
+		}
+		p.msgQueue.processPendingCommitMessages(p.state.view.Copy(), p.verifyCommitMessage)
```
ferranbt, line 803: I would imagine that the `msgQueue` should figure out if there is any new value in `pendingCommitQueue` and inject into the verification workflow if necessary.

ferranbt (changes requested):

consensus_pbft.go
Stefan-Ethernal: Think we can, although wanted to validate commit message as early as possible (which is as soon as we set the proposal).

msg_queue.go
Stefan-Ethernal: I see that if the round changes, we would discard those messages, as they would be from the previous round(s). That's the main benefit and rationale for using `msgQueueImpl`. So I'd say we shouldn't be validating commit messages against the wrong proposal (i.e. the one being from not from the current sequence and round). 
However I want to verify it with additional logs.

consensus_pbft.go
Stefan-Ethernal: Fixed https://github.com/0xPolygon/pbft-consensus/pull/78/commits/73c4d081b0a48e2c5a74773f783e9042f5d4ac2e

msg_queue.go
Stefan-Ethernal: Yes, reusing its logic and not having it at two places, since we are using it to read from both `pendingCommitMsgsQueue` (which is not related to any specific PBFT state) and from other queues (related to PBFT states, such as `validateStateQueue`, `roundChangeStatteQueue` etc.).

I was thinking initially of reusing `validateStateQueue` for pending commit messages as well, but that would require expanding commit messages with additional flag indicating its state (pending/non-validated vs confirmed/validated). Documented [here](https://www.notion.so/RFC-28-0e69004ef82241af88b16e41d51a15c2#86e6cc3ffd4444919d34087d78e4d8be).

msg_queue.go
igorcrevar: I think we have problem here. `PushMessage` used to be very simple, it used to just add new message to queue. With latest changes we introduced access to `view`, `proposal` and `velidatorset` from state inside method. And after that we have problem because those fields are accessed from more than one `go routine`.

ferranbt: I agree with @igorcrevar. The logic is really convoluted. It is hard to keep a mental model of the lifecycle of a commit message.

msg_queue.go
ferranbt: This is going to be a problem later on. This function is not only doing the discard of messages but doing internal logic (i.e RoundChangeState condition). You do not need that kind of logic when dealing with pending commit messages.

msg_queue.go
Stefan-Ethernal: What @igorcrevar was trying to explain is that we would need to have lock when accessing the state reference. That is because we are accessing to the state from both `PushMessage`, which is executed in separate routine. We need to do that, because message queue without the pbft state isn't able to determine on its own whether given commit message is valid. We need proposal hash, current view (sequence and round) and validator set, which all are defined in the state.

Not sure if we can avoid locking the state even if we go with your approach, since we are going to be accessing to it from multiple routines eventually. 🤔

msg_queue.go
Stefan-Ethernal: > This is going to be a problem later on.

I don't see this part of the code is going to be changed ever.

> You do not need that kind of logic when dealing with pending commit messages.

Ok agree with that part. In that case, I'd suggest introducing additional parameter which should handle message differently, based on given `PbftState` value.

However, please don't let we find ourselves forking logic and copying almost same functions multiple times, just because they aren't 100% the same.

Stefan-Ethernal:
Parallel commit messages validation is going to be delivered through #80. Closing this PR.

PR #80, opened 2022-08-16
Parallel validation commit messages

Stefan-Ethernal:
Validate commit messages in concurrently (in a separate go routine), before injecting those to the validate state queue. Only valid commit messages are pushed to the validate state queue.

Points to discuss:
- [x] Figure out how and when to call close function on commit validation routine https://github.com/0xPolygon/pbft-consensus/blob/77fe79f9b69fb5446218bea92b5aee9a740a0d09/msg_queue.go#L194
Tried invoking it in `Pbft.Run` function, but that won't work for e2e tests, which are reusing the same instance of PBFT when restarting a nodes (thus we end up in closing the same channel multiple times, which causes panic). https://github.com/0xPolygon/pbft-consensus/blob/77fe79f9b69fb5446218bea92b5aee9a740a0d09/consensus_pbft.go#L169
Most recent approach is introducing `Close` function to the `Pbft`, which invokes cancellation of commit validation routine. It should be called from `Polybft.Close` function on the client.
- [ ] TBD: Report Open Telemetry tracing for pending and invalid commit messages (if needed).

### DIFF
--- a/consensus_test.go
+++ b/consensus_test.go
@@ -756,6 +756,7 @@ func TestRoundChange_PropertyMajorityOfVotingPowerAggreement(t *testing.T) {
 				return nil
 			},
 		}, WithLogger(log.New(io.Discard, "", log.LstdFlags)))
+		defer node.Close()
 		node.state.validators = validatorSet
 		node.state.view = &View{
 			Round:    1,
--- a/consensus_test.go
+++ b/consensus_test.go
@@ -459,28 +459,20 @@ func TestTransition_ValidateState_MoveToCommitState(t *testing.T) {
 		m := newMockPbft(t, []NodeID{"A", "B", "C", "D"}, nil, "A")
 		m.setState(ValidateState)
 
-		// Prepare messages
-		m.emitMsg(createMessage(NodeID("A"), MessageReq_Prepare, nil))
-		m.emitMsg(createMessage(NodeID("B"), MessageReq_Prepare, nil))
-		// repeated message is not included
-		m.emitMsg(createMessage(NodeID("B"), MessageReq_Prepare, nil))
-		m.emitMsg(createMessage(NodeID("C"), MessageReq_Prepare, nil))
-
 		// Commit messages
 		m.emitMsg(createMessage(NodeID("B"), MessageReq_Commit, nil))
 		m.emitMsg(createMessage(NodeID("C"), MessageReq_Commit, nil))
+		m.emitMsg(createMessage(NodeID("D"), MessageReq_Commit, nil))
 
 		m.runCycle(context.Background())
 
 		m.expect(expectResult{
-			sequence:               1,
-			state:                  CommitState,
-			prepareMsgs:            3,
-			commitMsgs:             3, // Commit messages (A proposer sent commit via state machine loop, C and D sent commit via emit message)
-			commitMsgsVotingPower:  3,
-			locked:                 true,
-			outgoing:               1, // A commit message
-			prepareMsgsVotingPower: 3,
+			sequence:              1,
+			state:                 CommitState,
+			commitMsgs:            3, // Commit messages (B, C and D sent commit via emit message)
+			commitMsgsVotingPower: 3,
+			locked:                true,
+			outgoing:              1, // A commit message
 		})
 	})
 
@@ -499,8 +491,7 @@ func TestTransition_ValidateState_MoveToCommitState(t *testing.T) {
 		m.expect(expectResult{
 			sequence:              1,
 			state:                 CommitState,
-			prepareMsgs:           0,
-			commitMsgs:            3, // Commit messages (A proposer sent commit via state machine loop, C and D sent commit via emit message)
+			commitMsgs:            3, // Commit messages (B, C and D sent commit via emit message)
 			locked:                true,
 			outgoing:              1,  // A commit message
 			commitMsgsVotingPower: 85, // B. C and D sends commit messagess
@@ -571,7 +562,7 @@ func TestTransition_ValidateState_WrongMessageType(t *testing.T) {
 
 	// Create preprepare message and push it to validate state message queue
 	msg := createMessage(NodeID("A"), MessageReq_Preprepare, ViewMsg(1, 0))
-	heap.Push(&m.msgQueue.validateStateQueue, msg)
+	heap.Push(m.msgQueue.validateStateQueue, msg)
 	assert.PanicsWithError(t, "BUG: Unexpected message type: Preprepare in ValidateState from node A", func() { m.runCycle(context.Background()) })
 }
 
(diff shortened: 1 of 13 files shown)


msg_queue.go
```diff
@@ -34,6 +95,36 @@ func (m *msgQueue) readMessage(st State, current *View) *MessageReq {
 	return msg
 }
 
+func readValues(queue *msgQueueImpl, current *View) (*MessageReq, []*MessageReq) {
```
Stefan-Ethernal, line 98: Just because of that single if/else branch which checks `PbftState` we are duplicating rest of the code. 
Not a fan of this approach, to be honest and to me it is not in line with good coding practice.

Instead I think we should extract it into the function and have different handler defined for `RoundChangeState` and other `PbftState` values.

msg_queue.go
```diff
@@ -8,17 +8,78 @@ import (
...
+
+func (c *commitValidationRoutine) init() {
+	var round *RoundInfo2
+
+START:
+	for {
+		// read as many values as possible for the current view
+		val, _ := readValues(c.queue, round.view)
+		if val != nil {
+			// process in parallel
```
Stefan-Ethernal, line 40: We would need reference to state here (namely proposal hash, validator set and current view). Think that we would need a locking mechanism for the state, since we are accessing from multiple routines.

Stefan-Ethernal (commented):
Here is some of the feedback. However this approach still doesn't addresses the main pain point we are having at the moment, which is concurrent access to the `state` reference (namely validator set, current view, proposal hash).

msg_queue.go
```diff
@@ -130,21 +228,30 @@ func stateToMsg(st State) MsgType {
 	}
 }
 
-type msgQueueImpl []*MessageReq
+type msgQueueImpl struct {
+	queue    []*MessageReq
+	updateCh chan struct{}
```
Stefan-Ethernal, line 233: Is this the same channel we have channel which we have in the `Pbft`? If it is, then it may be some issues, because we are reading it in the `getNextMessage` and here in `msg_queue` in `init` function.

msg_queue.go
```diff
@@ -8,17 +8,78 @@ import (
...
+
+type RoundInfo2 struct {
+	Proposal []byte
+
+	// this is more static, it might be better to create it from another place
+	Validators ValidatorSet
+	view       *View
+}
+
+func (m *msgQueue) updateRound(info *RoundInfo2) {
```
Stefan-Ethernal, line 78: Are we going to call this from the beggining of `runValidateState`?

msg_queue.go
```diff
@@ -34,6 +95,36 @@ func (m *msgQueue) readMessage(st State, current *View) *MessageReq {
 	return msg
 }
 
+func readValues(queue *msgQueueImpl, current *View) (*MessageReq, []*MessageReq) {
+	discarded := []*MessageReq{}
```
Stefan-Ethernal, line 99: If we are going to keep this function, keep in mind that we need lock here as well, because `PushMessage` is invoked from a separate routine.

msg_queue.go
```diff
@@ -8,17 +8,78 @@ import (
...
 
+type commitValidationRoutine struct {
+	updateCh chan *RoundInfo2
+	closeCh  chan struct{}
+	queue    *msgQueueImpl
+}
+
+func (c *commitValidationRoutine) init() {
+	var round *RoundInfo2
+
```
ferranbt, line 34: workCh := make(chan Work)
worker := func() {
   for _, work := <-workCh {
     ...
   }
}

for i := 0; i < 5; i++ {
  go worker()
}

msg_queue.go
```diff
@@ -159,21 +266,21 @@ func (m msgQueueImpl) Less(i, j int) bool {
...
 // Swap swaps the places of the items at the passed-in indexes
 func (m msgQueueImpl) Swap(i, j int) {
-	m[i], m[j] = m[j], m[i]
+	m.queue[i], m.queue[j] = m.queue[j], m.queue[i]
 }
 
 // Push adds a new item to the queue
 func (m *msgQueueImpl) Push(x interface{}) {
-	*m = append(*m, x.(*MessageReq))
+	m.queue = append(m.queue, x.(*MessageReq))
```
ferranbt, line 274: select {
  case m.updateCh <- struct{}:
  default:
}

msg_queue.go
```diff
@@ -35,69 +53,161 @@ func (m *msgQueue) readMessage(st State, current *View) *MessageReq {
...
+	}
+	// validateMsgWorker validates single commit message
+	validateMsgWorker := func(commitMsgsCh <-chan *MessageReq, stateInfo *stateInfo) {
+		for commitMsg := range commitMsgsCh {
+			if err := stateInfo.validators.VerifySeal(commitMsg.From, commitMsg.Seal, stateInfo.proposalHash); err != nil {
+				doneCh <- struct{}{}
+				c.logger.Printf("[ERROR] Commit message is invalid (%v). Error: %s", commitMsg, err)
+				return
+			}
+			doneCh <- struct{}{}
```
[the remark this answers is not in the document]
Stefan-Ethernal: Purpose of `doneCh` is to block until all messages are processed and I need it because of unit test (https://github.com/0xPolygon/pbft-consensus/pull/80/files#diff-353efbbf5c50447329b1d508abf3f0d44552fd09f810ba1364d67bf5d7ef7f08R165). However even with this change, unit test is flaky for some reason.

Do you have any idea how to figure out when `commitValidationRoutine` is done? Tricky part to me is that we can have more messages than workers and therefore not sure what is the best pattern to figure out when all the messages we have pushed into the queue are processed...? :thinking:

msg_queue.go
```diff
@@ -35,69 +53,161 @@ func (m *msgQueue) readMessage(st State, current *View) *MessageReq {
...
+
+		for {
+			c.lock.RLock()
+			// read as many messages as possible for the current view
+			commitMsg, _ := c.pendingCommitMsgs.readMessageWithDiscardsLocked(stateInfo.view, ValidateState)
+			c.lock.RUnlock()
+			if commitMsg == nil {
+				break
+			}
+			commitMsgsCh <- commitMsg
```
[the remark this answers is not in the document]
Stefan-Ethernal: I thought that it is enough to close `commitMsgsCh` which would immediately interrupt further processing of given messages: https://github.com/0xPolygon/pbft-consensus/pull/80/files#diff-4fc32a02f0b1e925a6cca81c55e06a1adca2df98069efd502785727ea647ff77R169-R172.

Any suggestions on alternative approach?

consensus_test.go
```diff
@@ -459,28 +459,20 @@ func TestTransition_ValidateState_MoveToCommitState(t *testing.T) {
...
 
 		m.expect(expectResult{
-			sequence:               1,
-			state:                  CommitState,
-			prepareMsgs:            3,
-			commitMsgs:             3, // Commit messages (A proposer sent commit via state machine loop, C and D sent commit via emit message)
-			commitMsgsVotingPower:  3,
-			locked:                 true,
-			outgoing:               1, // A commit message
-			prepareMsgsVotingPower: 3,
```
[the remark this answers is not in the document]
Stefan-Ethernal: Yes, this is also due to the fact that messages are processed in parallel and we don't (yet) have mechanism to wait for all the emitted messages get processed until we make assertions. 

Therefore most of the time manages to process `commit` messages before the `prepare` messages (since `commit` message type takes precedence over `prepare` if both are for the same sequence and round, because of `iota` value and message queue sorting logic).
So that's why I've opted to emit just a commit messages and bypass the problem.

msg_queue.go
```diff
@@ -2,21 +2,29 @@ package pbft
...
+	"math"
 	"sync"
 )
 
 // msgQueue defines the structure that holds message queues for different PBFT states
 type msgQueue struct {
-	// Heap implementation for the round change message queue
-	roundChangeStateQueue msgQueueImpl
+	// roundChangeStateQueue is a heap implementation for the round change message queue
+	roundChangeStateQueue *msgQueueImpl
```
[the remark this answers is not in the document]
Stefan-Ethernal: I'm ok with either approach. In fact think that you have changed this into the pointer. :slightly_smiling_face:

msg_queue.go
```diff
@@ -35,69 +53,161 @@ func (m *msgQueue) readMessage(st State, current *View) *MessageReq {
...
+		case info := <-c.updateStateInfoCh:
+			stateInfo = info
+			close(commitMsgsCh)
+			commitMsgsCh = make(chan *MessageReq, pendingCommitsBufferSize)
+		case <-c.closeCh:
+			return
+		}
+
+		c.lock.RLock()
+		workersNum := int(math.Min(float64(maxWorkersCount), float64(c.pendingCommitMsgs.Len())))
```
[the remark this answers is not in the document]
Stefan-Ethernal: So if I understand correctly, just need to move this part out of the `for` loop, always create `maxWorkersCount` and that should be all?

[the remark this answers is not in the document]
Stefan-Ethernal: At first I was thinking that it should be enough just moving spawning of routines above of given for loop, however apparently implementing it this way, we end up not invoking `validateMsgWorker` function ever... Any ideas on this?

```go
func (c *commitValidationRoutine) run(doneCh chan struct{}) {
	if doneCh == nil {
		doneCh = make(chan struct{})
	}
	// validateMsgWorker validates single commit message
	validateMsgWorker := func(commitMsgsCh <-chan *MessageReq, stateInfo *stateInfo) {
		for commitMsg := range commitMsgsCh {
			if err := stateInfo.validators.VerifySeal(commitMsg.From, commitMsg.Seal, stateInfo.proposalHash); err != nil {
				c.logger.Printf("[ERROR] Commit message is invalid (%v). Error: %s", commitMsg, err)
				doneCh <- struct{}{}
				return
			}
			doneCh <- struct{}{}
			c.validMsgHandler(commitMsg)
		}
	}

	var stateInfo *stateInfo
	var commitMsgsCh chan *MessageReq
	for i := 0; i < maxWorkersCount; i++ {
		go validateMsgWorker(commitMsgsCh, stateInfo)
	}

	for {
		// wait for:
		// 1. new messages on the queue
		// 2. update of the round
		// 3. close trigger
		select {
		case <-c.notifyPendingCommitCh:
			if stateInfo == nil {
				continue
			}
		case info := <-c.updateStateInfoCh:
			stateInfo = info
			if commitMsgsCh != nil {
				close(commitMsgsCh)
			}
			commitMsgsCh = make(chan *MessageReq, pendingCommitsBufferSize)
		case <-c.closeCh:
			return
		}

		for {
			c.lock.RLock()
			// read as many messages as possible for the current view
			commitMsg, _ := c.pendingCommitMsgs.readMessageWithDiscardsLocked(stateInfo.view, ValidateState)
			c.lock.RUnlock()
			if commitMsg == nil {
				break
			}
			commitMsgsCh <- commitMsg
		}
	}
}
```

vcastellm:
Can we close this?

Stefan-Ethernal:
> Can we close this?

Yes I would say so.

PR #87, opened 2022-08-22
Fix order in validate

ferranbt:
This PR changes the order in the validation of preprepare proposals. Before, the `Backend.Validate` function would be called before checking if the locked proposal was valid. Thus, the internal state of the backend could be changed for a state that was invalid.


vcastellm (approved):
LGTM

b00ris (approved):

Stefan-Ethernal (approved):

PR #88, opened 2022-08-23
Move unlocking when setting sequence

Stefan-Ethernal:
If node gets out of sync at some point and it previously locked some proposal, it won't unlock it even if it manages to sync up with network and proceed to the current sequence.

Therefore we are moving unlocking logic from the `runCommitState` to `setSequence` function (which is being executed at the beginning of the consensus algorithm).


vcastellm (approved):
LGTM

